PR #1, opened 2016-04-12
make css entries configurable

matthias-reis:
This pr gets rid of the direct dependency to "normalize.css" and instead makes this part configurable by the application.

Simply pass the dependencies as an array in `.hopsrc.js`, key: `cssAdditionalEntry`, default: `[]`

### DIFF
--- a/etc/webpack.base.js
+++ b/etc/webpack.base.js
@@ -5,7 +5,7 @@ var config = require('../lib/config');
 
 module.exports = new WebpackConfig().merge({
   entry: [
-    'normalize.css',
+    ...config.cssAdditionalEntry,
     config.appRoot
   ],
   output: {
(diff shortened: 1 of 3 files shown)


etc/webpack.base.js
```diff
@@ -5,7 +5,7 @@ var config = require('../lib/config');
 
 module.exports = new WebpackConfig().merge({
   entry: [
-    'normalize.css',
+    ...config.cssAdditionalEntry,
```
dmbch: I think this should be ES5 - I think we should stick to vanilla ES5 for core (for now)

dmbch:
...and would you mind adding yourself to the contributors list in package.json? :smile:

dmbch:
Thanks a lot!

PR #2, opened 2016-04-12
Adds postinstall script

rcsole:
Adds a postinstall script which will try to create some boilerplate, so you can easily
do something like:

```sh
mkdir my-app && cd my-app
npm init && npm i hops -SE
npm start
```

And just have it work.

### DIFF
--- a/bin/setup.js
+++ b/bin/setup.js
@@ -1,0 +1,52 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+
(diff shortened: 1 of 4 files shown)


dmbch:
Thanks a lot! I think this makes a lot of sense. I have one really minor request, though: Would you mind limiting line length (in js files) to 80 characters?

Other than that, I'll be really happy to merge.

rcsole:
Done!

dmbch:
Great, thanks!

PR #3, opened 2016-04-12
Adds postcss-modules-values

rudeworks:
Adds a dependency for the package [postcss-modules-values](https://www.npmjs.com/package/postcss-modules-values) so the user can define and share variables between components inside the app.

**Open question:** 

Would it be good to add more of the logic this package offers to the example app so we can also test the css output?

### DIFF
--- a/app/src/style.css
+++ b/app/src/style.css
@@ -1,4 +1,6 @@
 
+@value primary: crimson;
+
 .headline {
-  color: crimson;
+  color: primary;
 }
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "postcss": "5.0.19",
     "postcss-cssnext": "2.5.2",
     "postcss-loader": "0.8.2",
+    "postcss-modules-values": "1.1.2",
     "react": "15.0.1",
     "react-addons-update": "15.0.1",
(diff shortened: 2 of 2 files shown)


dmbch:
Thank you very much!

Also, I think it would really make sense to add way more css examples, similar to what we did for js: I think app/main.js already contains most of what's interesting about the js side of the stack, while style.css still is almost empty. I think I'd like to show off some of the CSSNext features (like nesting etc.), module composition...

rudeworks:
@dmbch I'll prepare another PR with that postcss-next feature examples. postcss-modules-values also let's you import @values from external files and I think it would be cool to show that in the example. I would create another file in the src folder called *values.css* and then use them to store example variables like media query definitions and colors, what do you think?

rcsole:
@dmbch @rudeworks I think it would be useful to leave `app/src` as simple
as we can. Given that since #2 it's being used as a template. Maybe we can
add an `examples` folder?

rudeworks:
@rcsole @dmbch I don't think there's that many examples still left to be added, so maybe an examples folder that you have to erase every time you start a new project is actually adding complexity. I guess it's enough with the app folder and that extra css file.

PR #8, opened 2016-04-14
Enables Redux DevTools Chrome extension

### LINKED ISSUE #7: Redux DevTools can't find store
The store enhancer is missing some configuration:

https://github.com/zalmoxisus/redux-devtools-extension#2-use-with-redux

rcsole:
Redux DevTools was not being able to access the store because we needed
to set it up. This closes #7.

### DIFF
--- a/lib/store.js
+++ b/lib/store.js
@@ -9,6 +9,15 @@ function isFunc(object) {
   return (typeof object === 'function');
 }
 
+function devTools () {
+  if (typeof window === 'object'
+      && typeof window.devToolsExtension !== 'undefined') {
+    return window.devToolsExtension();
+  } else {
+    return function (f) { return f; };
+  }
+}
+
 module.exports = {
   createReducer: function (key) {
     var type = 'update' + key.charAt(0).toUpperCase() + key.slice(1);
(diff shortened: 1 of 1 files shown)


lib/store.js
```diff
@@ -9,6 +9,15 @@ function isFunc(object) {
   return (typeof object === 'function');
 }
 
+function devTools () {
+  if (typeof window === 'object'
+      && typeof window.devToolsExtension !== 'undefined') {
```
matthias-reis: Just a question: Do we want the devTools even in production mode?

rcsole: Probably not, let me fix that.

dmbch: Why wouldn't we, actually? I really kind of like the idea to have these kinds of debugging capabilities production, the more so as having them wouldn't even bloat our bundle...

rcsole: Hmmm... I guess you're right. Worst that can happen is someone
hooks on our store with their DevTools, it's not like they access the
store itself so it should be fine.

rcsole: cc @rudeworks @naoisegolden

rcsole:
I've re-enabled it for production again, I do think it would be useful.

dmbch:
Thank you! :+1:

PR #9, opened 2016-04-14
Annotates template source code

### LINKED ISSUE #5: Annotated source
It might be a good idea to provide an annotated source in the example
app, instead of it being plain. Eg. comment what the decorator does,
`getSelector`, `createReducer`, etc...

Along with maybe a list of useful resources that might explain more in-depth
the tools and libraries this bundle includes.

Thoughts?

rcsole:
We still need to discuss if we want to talk about "stateless" component
in the first part. Any suggestions?

Closes #5.

### DIFF
--- a/app/src/main.js
+++ b/app/src/main.js
@@ -5,6 +5,13 @@ import { connect } from 'react-redux';
 
 import { createReducer, render } from 'hops';
 
+/*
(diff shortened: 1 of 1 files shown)


dmbch:
Thank you!

Let's see if we can come up with a nice intro to stateless components later - I'll see if I can give it a shot shortly.

PR #13, opened 2016-04-19
Adds the .pcss extension to the CSS loader test

rcsole:
I personally prefer `.pcss` extension ever since @rudeworks showed it to
me, as it gives me cool syntax highlighting in some editors (eg. VS Code).

### DIFF
--- a/etc/webpack.build.js
+++ b/etc/webpack.build.js
@@ -11,7 +11,7 @@ var baseConfig = path.join(__dirname, 'webpack.base');
 module.exports = new WebpackConfig().extend(baseConfig).merge({
   module: {
     loaders: [{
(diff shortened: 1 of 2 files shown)


rcsole:
It does not seem to be working, I'll investigate and re-open.

rudeworks:
Actually, I started using plain ```.css``` files since I'm just using ```postcss-cssnext``` which let's you write valid CSS (or at least future valid CSS). It's basically the same as writing ES2015 javascript, you don't name files differently just because you're using a future syntax.

Isn't VS Code able to change syntax highlighting on a single file? I just chose PostCSS syntax highlighting on ```.css``` files in Atom and it works beautifully.

PR #19, opened 2016-04-20
Components

dmbch:
This pr makes hops (and hops based products) strictly ES2015 compliant: no more decorators, no more class properties. It also introduces a couple of new helper functions (`createContainer()`, `createComponent()`).


rcsole:
It seems like this PR is also removing the example of importing values
in CSS, is that intended?

dmbch:
Yeah, it is: the additional css file was not copied by the setup script, thus breaking package installation. And I wasn't keen on complicating that script further.

I was hoping that with the comments/examples, style.css would be enough to show most of the css possibilities...

Btw, this PR still needs some work anyway.

matthias-reis:
The new syntax looks good to me. I'll go for it in `malt` ... I think we need a more extensive description of internal coding guidelines though, because this is not opinionated anymore (which is a good thing by the way).

naoisegolden:
@matthias-reis can you tell more about `malt`?

matthias-reis:
Yes, of course. `malt` should become the internal counterpart of hops and acts like e.g. twitter bootstrap - or like a newer version of our xing patterns.

As Ricard pointed out, this is done in one repo `frontend/malt`which acts like a babel inspired monorepo. Any pattern is a very lightweight ES6 React Component, holding its own dependencies. So all users (= developers) can use it partially - e.g only the grid, only the frame, only header and footer of the frame and so on.

There will also be some special packages that are worth mentioning:
- `@malt/predicates` will simply be a collection of xing css variables
- `@malt/basestyles` will be initial global styling (providing normalization and styling `body`, `h1`-`h6`, `p` ...)
- `@malt/shell` will provide a component that renders the whole frame for you (header, navigation and footer like `<XingShell>[... your stuff ...]</XingShell>`)
- `@malt/svg` (not yet started) will provide you with an svg sprite solution (also used as a successor of the iconfont) 
- `@malt/i18n` will provide connection to our phrase app output and other internationalization tools

I'm sure there's more to come ... 
If you have questions or more ideas that go into that direction, please contact me or contribute with pull requests.

This week my personal goal is to get the publishing process ready so you will be able to use these packages together with our internal npm repo.

matthias-reis:
Hmmm ... I'm going to use parts of that in the readme.md :-) @naoisegolden Thanks for pointing to the missing docs ...

naoisegolden:
@matthias-reis sounds really good, thanks.

PR #24, opened 2016-04-25
Renames 'noBabel' config option to a more descriptive 'babelIgnore'

rudeworks:
```noBabel``` sounds like an option requiring a boolean, I'm proposing ```babelIgnore``` as a more descriptive option name.

### DIFF
--- a/etc/webpack.base.js
+++ b/etc/webpack.base.js
@@ -16,7 +16,8 @@ module.exports = new WebpackConfig().merge({
   module: {
     loaders: [{
       test: /\.js$/,
-      loader: 'babel', exclude: config.noBabel
+      loader: 'babel',
+      exclude: config.BabelIgnore
     }, {
       test: /\.json$/,
       loader: 'json'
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: etc/webpack.base.js [hunk 1]
```diff
@@ -16,7 +16,8 @@ module.exports = new WebpackConfig().merge({
   module: {
     loaders: [{
       test: /\.js$/,
-      loader: 'babel', exclude: config.noBabel
+      loader: 'babel',
+      exclude: config.BabelIgnore
```
dmbch: I think it ought to be `config.babelIgnore` ;)

lib/render.js
```diff
@@ -2,7 +2,7 @@
 var config = require('../lib/config');
 
 require('babel-register')({
-  ignore: config.noBabel
+  ignore: config.BabelIgnore
```
dmbch: See above.

dmbch:
Thanks! This is way better semantically.

etc/webpack.base.js
rudeworks: fixed 🐕💨

PR #30, opened 2016-04-26
add testing support using tape/enzyme

dmbch:


rcsole:
👍 it seems like I'm removing Jest from my project!

PR #31, opened 2016-04-26
Logs possible errors on server-side render function

rcsole:
I kept getting stuff broken during server side rendering and it wouldn't
log anything out. This makes server-side rendering log errors as well.


dmbch:
Thank you!

PR #60, opened 2016-06-07
Adding support for jsx files

drixsonic:
Adding support for jsx files for the ones who would like to have it.


dmbch:
Thanks, mate! 👍

PR #101, opened 2016-08-17
template generation with asset path prefix

fdietz:
Prefix all script and stylesheet paths with asset path prefix based on webpack output.publicPath config.


dmbch:
Thanks a lot! ✨

PR #102, opened 2016-08-18
output.publicPath set to http://localhost:8080/

fdietz:
see http://webpack.github.io/docs/webpack-dev-server.html#combining-with-an-existing-server docs 

This enables the use of another backend server (for example Ruby on Rails) to serve the Webpack generated index.html template which itself references assets from http://localhost:8080.


dmbch:
👋 as discussed irl, I'll close this for no.

Thanks, tho!

PR #153, opened 2017-03-08
refactor plugin

dmbch:
refactor plugin, replace custom code with existing npm modules

PTAL @robin-drexler


dmbch:
Not good enough, yet.

robin-drexler:
Oh, was about to start a review.

PR #157, opened 2017-04-04
Release 5.0.0

dmbch:
refactoring, removal of `close` method


dmbch:
PTAL, @robin-drexler: I tried to a) streamline that transpiler thingy and b) improve its legibility/grokability. What do you think?

PR #164, opened 2017-06-15
Allow dynamic imports for node as well

robin-drexler:


robin-drexler:
https://github.com/xing/hops/pull/164/files#diff-d32b7bc6606ed8db9edcbf12f8670e84R47 is the interesting part :)

PR #175, opened 2017-06-28
docs(hops-cli): add readme

systemboogie:
This adds a readme to the hops-cli package.

### DIFF
--- a/packages/cli/readme.md
+++ b/packages/cli/readme.md
@@ -4,7 +4,7 @@ hops-cli provides a small set of commands to manage your hops project. Installin
 
 # Installation
 ## Minimum installation for any project
-We assume that you already created a folder for your project and initialized it with `npm init`. Go to the root folder of your project and install hops-cli. This installs hops-config and hops-server as dependencies as well:
+We assume that you already created a folder for your project and initialized it with `npm init -y`. Go to the root folder of your project and install hops-cli. This installs hops-config and hops-server as dependencies as well:
 ``` bash
 npm install --save hops-cli
 ```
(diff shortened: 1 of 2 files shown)


packages/cli/readme.md
```diff
@@ -1,4 +1,79 @@
...
+
+``` JSON
+"scripts": {
+    "build": "hops build",
+    "develop": "hops develop",
+    "serve": "hops serve",
+    "start": "hops start"
+  }
+```
+
```
dmbch: I'd probably add that you might only need to add `"start": "hops start"` for most projects.

packages/cli/readme.md
```diff
@@ -1,4 +1,79 @@
...
+``` bash
+npm run build
+npm run develop
+npm run serve
+npm start
+npm start --production 
+```
+
+## Very basic example app
+To try out the [minimum installation described above](#minimum-installation-for-any-project), create a main.js file in the root folder of your project and put the following code in there:
```
dmbch: Why don't we go for `index.js` here? That's the default `npm init -y` configures and by using that, we wouldn't have to explain how to change the `main` field...

dmbch:
Great stuff, by the way! Thank you very much!

packages/cli/readme.md
systemboogie: Good point! Added that.

packages/cli/readme.md
systemboogie: I'm not much of an `npm init -y` guy, because I like to set some properties upfront.  But as we're in context of an example app here, it's much easier taking that route. Changed it, PTAL 👀

PR #191, opened 2017-07-20
docs(hops-config): add readme

systemboogie:
This adds a readme to the hops-config package.


packages/config/readme.md
```diff
@@ -1,4 +1,126 @@
-
 # Hops Config
 
-T.B.D. (To Be Documented)
+hops-config exposes options to configure the other Hops packages and Webpack.
+
+# Installation
+When you install [hops-cli](https://www.npmjs.com/package/hops-cli), hops-config will be installed as a dependency as well - so normally, you don't have to separately install hops-config (and we don't recommend that). If you for some reason need to do that anyways, just do:
```
dmbch: I think we _do_ recommend installing `hops-config` manually iff you want to import/require it in downstream code.

packages/config/readme.md
```diff
@@ -1,4 +1,126 @@
...
+
+```
+npm install --save hops-config
+```
+
+# Usage
+
+Please find a list of all available options below. They can be set in your project's package.json and can be overridden by using [npm config](https://docs.npmjs.com/cli/config) in the command line. It's important to understand that the configuration is immutable at runtime, i.e. you can't for example import values and change them. hops-config provides an `extends` field as a means for augmenting the existing configuration.
```
dmbch: I think I'd get rid of the half sentence starting with "i.e." - I think they're a bit confusing, to be honest.

Additionally, we might want to address the fact that `hops-config` can not just be extended, but also imported in downstream projects: its isomorphic importability is something that I rather like. What do you think?

packages/config/readme.md
```diff
@@ -1,4 +1,126 @@
...
+| `appDir` | `String` | Directory of your app. This way, you can put your app to a place other than project root. Defaults to `"."` |
+| `buildDir` | `String` | Directory where the build artifacts (i.e. your application/asset files) will be saved to. Defaults to `"build"`. Hops always removes the specified directory before starting a new build |
+| `cacheDir` | `String` | Directory where internally used artifacts not meant to be served will be output (e.g. the `manifest.json`/`manifest.js` files generated by Webpack). Defaults to `"node_modules/.cache/hops"` |
+| `buildConfig` | `String` | Path to your Webpack build configuration file. Defaults to Defaults to `./node_modules/hops-config/configs/build.js` |
```
dmbch: Double "Defaults to" :)

packages/config/readme.md
```diff
@@ -1,4 +1,126 @@
...
+| `developConfig` | `String` | Path to your Webpack development configuration file. Defaults to `./node_modules/hops-config/configs/develop.js` |
+| `nodeConfig` | `String` | Path to your Webpack node/server-side rendering configuration file. Defaults to `./node_modules/hops-config/configs/node.js` |
+| `manifest` | `String` | The application's `manifest.js` as a string. Defaults to a getter function that returns the content of the automatically generated `manifest.js` |
+| `assets` | `Object literal` | The applications's `.js` and `.css` assets. Defaults to a getter function that returns the assets contained in the automatically generated `manifest.json` |
+| `extends` | `String` | Path to a baseline configuration (could be for example a node module or a project-specific file), allowing you join the current configuration with that baseline |
+
+hops-config normalizes all specified directory and path strings. That means the following statements yield the same result:
```
dmbch: I'm afraid we're mixing up two separate features here: normalization happens for some HTTP path related values (`locations`, `basePath`, `assetPath`) and is something I probably wouldn't even mention here.

On the other hand, everything matching `/(config|file|dir)s?$/i` will be treated as a filesystem path (or array thereof) and resolved relative to app root. This only applies if the respective value is not an absolute path already.

packages/config/readme.md
systemboogie: Oh yes :-)

packages/config/readme.md
systemboogie: > I think I'd get rid of the half sentence starting with "i.e."

Hmm... I don't think it's confusing, but I'm not in love with that sentence anyways :-)

> Additionally, we might want to address the fact that `hops-config` can not just be extended, but also imported in downstream projects: its isomorphic importability is something that I rather like. What do you think?

Fair point, added that. This should definitely be mentioned in the top-level documentation as well :-)

packages/config/readme.md
systemboogie: > I'm afraid we're mixing up two separate features here: normalization happens for some HTTP path related values (`locations`, `basePath`, `assetPath`) and is something I probably wouldn't even mention here.

Actually I had that correctly in a previous version already with `basePath`. But anyways, I learned that such sentence doesn't tell the complete story, because the behavior differs for each of these keys. So let's omit it.

> On the other hand, everything matching `/(config|file|dir)s?$/i` will be treated as a filesystem path (or array thereof) and resolved relative to app root. This only applies if the respective value is not an absolute path already.

Okay, let's add that to the page. One thing, though, when you say "relative to app root": Maybe we should say "relative to project root" in order to avoid people mixing up app root with appDir?

PR #197, opened 2017-07-28
feat(hops-jest-preset): add jest preset

robin-drexler:


packages/jest-preset/jest-preset.json
```diff
@@ -0,0 +1,13 @@
+{
+  "moduleNameMapper": {
+    "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/node_modules/jest-preset-hops/fileMock.js",
```
dmbch: I think we ought to support the same file types here and in `url`/`file` loaders. That would mean these:

`(png|gif|jpe?g|webp|html|svg|((o|t)tf)|woff2?|ico)`

packages/jest-preset/jest-preset.json
```diff
@@ -0,0 +1,13 @@
+{
+  "moduleNameMapper": {
```
dmbch: Don't we also have to mock `tpl` loader, i.e. provide a way to require a file with `tpl` extension and have a function returned?

packages/jest-preset/jest-preset.json
```diff
@@ -0,0 +1,13 @@
...
+    "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/node_modules/jest-preset-hops/fileMock.js",
+    "^.+\\.css$": "identity-obj-proxy"
+  },
+  "transform": {
+    "^.+\\.(js|jsx)$": "<rootDir>/node_modules/jest-preset-hops/transform.js"
```
dmbch: Aren't we missing `mjs` here?

packages/jest-preset/transform.js
```diff
@@ -0,0 +1,33 @@
+var babelJest = require('babel-jest');
+
+// eslint-disable-next-line import/no-internal-modules
```
dmbch: I'm not sure we need this in with semistandard :)

robin-drexler:
@dmbch addressed your findings, could you PTAL again? Cheers! :)

PR #198, opened 2017-07-28
Better integration tests

dmbch:


spec/develop.js
```diff
@@ -56,13 +56,15 @@ describe('development server', function () {
     })
     .then(function (body) {
       assert(body.length);
+      assert(body.indexOf('<!doctype html>') === -1);
```
robin-drexler: How about asserting in the previous `.then` that the `Content-Type` response header is `application/javascript` instead of `text/html` as well?

spec/serve.js
```diff
@@ -65,6 +52,45 @@ describe('production server', function () {
 
   it('should create build files', function () {
     var fileNames = fs.readdirSync(buildDir);
-    assert(fileNames.length = 2);
+    assert.equal(fileNames.length, 3);
```
robin-drexler: In order to understand those numbers, you have to have the mock application's structure and content in mind at all times. 

I think this could be improved by either adding an explanatory comment or changing the assertions to check the actual files.

spec/serve.js
```diff
@@ -65,6 +52,45 @@ describe('production server', function () {
...
+  });
+
+  it('should deliver expected html page', function (done) {
+    fetch('http://localhost:8080/')
+    .then(function (response) {
+      assert(response.ok);
```
robin-drexler: The content type response header could also be asserted to be `text/html`

spec/serve.js
```diff
@@ -65,6 +52,45 @@ describe('production server', function () {
...
+      return response.text();
+    })
+    .then(function (body) {
+      assert(body.length);
+      assert(body.indexOf('Hello World!') > -1);
+      done();
+    });
+  });
+
+  it('should deliver all asset files', function (done) {
```
robin-drexler: This is some integrative integration test! ❤️

spec/serve.js
```diff
@@ -65,6 +52,45 @@ describe('production server', function () {
...
+          assert.equal(body, fs.readFileSync(
+            path.resolve(
+              buildDir,
+              manifest[key].replace(/^\/?/, '')
+            ),
+            'utf8'
+          ));
+        });
+      })
+    ).then(function () { done(); });
```
robin-drexler: Instead of explicitly calling `done`, [according to the mocha docs](https://mochajs.org/#working-with-promises), you'd also just return the promise.

spec/serve.js
dmbch: Yes, you're right - but that's nothing hops does anything about, but rather an express default behavior... I'd postpone for now.

spec/develop.js
dmbch: Yeah, nice idea - though I'd like to postpone that, as it's not really about hops, but rather webpack-dev-server behaviour.

dmbch:
@robin-drexler PTAnotherL

demo/spec/home.spec.js
```diff
@@ -1,8 +1,14 @@
+/* eslint-env jest */
+
 import React from 'react';
 import renderer from 'react-test-renderer';
-import { Home } from '../src/Home';
+import ConnectedHome from '../src/home';
+
+const { WrappedComponent: Home } = ConnectedHome;
```
robin-drexler: Nice! Accessing the static property `WrappedComponent` instead of explicitly exporting the unconnected component is an interesting (not in the bad way) approach.

dmbch: Thanks! I thought so, too - doing so doesn't increase the exposed API surface...

PR #200, opened 2017-07-29
Updates & Cleanup

dmbch:


packages/jest-preset/transform.js
```diff
@@ -2,21 +2,6 @@ var babelJest = require('babel-jest');
 
 var hopsBabelConfig = require('hops-config/loaders/babel').node.use.options;
 
-// presets look like this
-// we try to figure out at which position "env" is and and set modules to "commonjs" to transpile import statements in jest
```
robin-drexler: I agree with removing the comment about the structure, however personally I'd leave the apart about why this is happening here (allowing import statements in jest)

packages/react/package.json
```diff
@@ -10,17 +16,14 @@
...
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: run tests in xing/hops repo\" && exit 1"
   },
-  "keywords": [],
-  "author": "",
-  "license": "MIT",
   "peerDependencies": {
     "hops-config": "6.1.0",
-    "react": "^15.5.4",
+    "react": "^15.6.1",
```
matthias-reis: Shouldn't we be more conservative with the peerDependencies and e.g. allow react down to 15.4?

dmbch: That would make sense - but, seeing that we only ever test on the latest versions of our dependencies, it would give us quite a blind spot. Obviously, we could simply rely on upstream's interpretation and implementation of semver... 🤷‍♂️

Let's discuss this irl.

PR #207, opened 2017-08-10
Release 6.2.1

dmbch:


KingHenne:
You should also revert my README change regarding the https config option.

packages/cli/lib/common.js
```diff
@@ -14,7 +14,7 @@ function defaultCallback (error) {
   } else {
     console.log('hops: Server listening at ' + url.format({
       protocol: hopsConfig.https ? 'https' : 'http',
-      hostname: hopsConfig.host,
+      hostname: hopsConfig.host === '0.0.0.0' ? 'localhost' : hopsConfig.host,
```
robin-drexler: Can you elaborate on why that is necessary?

packages/config/plugins/write-manifest.js
```diff
@@ -0,0 +1,35 @@
...
+module.exports = function Plugin (regExp) {
+  mkdirp.sync(hopsConfig.cacheDir);
+  this.apply = function (compiler) {
+    compiler.plugin('emit', function (compilation, callback) {
+      var paths = Object.keys(compilation.assets);
+      var assets = paths.reduce(function (assets, path) {
+        if (path.indexOf('hot-update') >= 0 || path.indexOf('chunk-') >= 0) {
+          return assets;
+        }
+        if (path.indexOf('.css') === path.length - 4) {
```
robin-drexler: `path` is a shadowed variable here which makes it harder to understand. 

Did you think about using `path.extname(assetPath) === '.css')`? That could make the intend a bit more obvious.

packages/config/plugins/write-manifest.js
```diff
@@ -0,0 +1,35 @@
...
+          if (path.indexOf('vendor-') >= 0) {
+            assets.js.unshift('/' + path);
+          } else {
+            assets.js.push('/' + path);
+          }
+        }
+        return assets;
+      }, { css: [], js: [] });
+      var fileName = path.resolve(hopsConfig.cacheDir, 'manifest.json');
+      var fileContent = JSON.stringify(assets, null, '  ');
```
robin-drexler: `JSON.stringify(assets, null, '  ');` could also be `JSON.stringify(assets, null, 2);`

packages/cli/lib/common.js
dmbch: I've been asked about this pseudo-IP a couple of times - while technically correct-ish, it does prevent users from copying/pasting the URL.

https://en.m.wikipedia.org/wiki/0.0.0.0

robin-drexler:
Only tried in demo project but it appears that now even the bug with missing assets on first load in dev mode is fixed. :)

Added some minor remarks, but no show stoppers from my perspective. :)

KingHenne:
I think you also need to adjust [this line](https://github.com/xing/hops/blob/master/packages/jest-preset/transform.js#L3). Maybe there's even more?

PR #209, opened 2017-08-16
(2/4) chore(redux): add flow-typed definitions for redux packages

ZauberNerd:
This commit adds type definitions for "redux" and "react-redux" from
the flow-typed repository to the "hops-redux" package, so that they
can be used by consumers of "hops-redux".

In order to use these definitions in a project that uses "hops-redux"
they need to be added to the project's ".flowconfig" "libs" section:
```
[libs]
node_modules/hops-redux/flow-typed/
```

These type definitions have been installed via the flow-typed CLI.
Read more about it here: https://github.com/flowtype/flow-typed


ZauberNerd:
Please merge in correct order to fix test failures:
1. #212 
2. #209 
3. #210 
4. #211

ZauberNerd:
Not needed anymore, as we decided to not put the type definitions inside hops, but rather into the projects that use hops. See #211

PR #213, opened 2017-08-29
update dependencies

dmbch:


package.json
```diff
@@ -8,7 +8,7 @@
     "postcss"
   ],
   "scripts": {
-    "start": "cd demo; rm -rf dist node_modules/*hops*; yarn install --force; yarn start",
+    "start": "cd demo; rm -rf dist node_modules/*hops*; npm install --no-package-lock; npm start",
```
robin-drexler: Why `--no-package-lock`? Asking because  there's been no `--no-lockfile` when yarn was used?

dmbch: I probably forgot to add it in the yarn case. Generally, I am trying here to force the package manager to fetch the freshest versions of all packages to simulate the behaviour in Travis (and in actual new projects).

dmbch:
@robin-drexler PTAnotherL :)

robin-drexler:
ja

PR #221, opened 2017-09-14
docs: add missing peerDependencies to hops-react/redux readme example

ZauberNerd:
The package `hops-react` has a peerDependency to `hops-config` but it
isn't specified in the package's `readme.md` installation instructions.

The package `hops-redux` has a peerDependency to `react` but it isn't
specified in the package's `readme.md` installation instructions.

This PR adds the missing peerDependencies to the installation
instructions in both readme files.


dmbch:
Nice catch! Thank you!

PR #222, opened 2017-09-14
fix(hops-config): allow config to be overridden by npm env variables

ZauberNerd:
With the PR #220 we lost the ability to override parts of the config
via npm environment variables `npm config set ...`.
By merging the environment config with the package config we can re-
introduce this behaviour.

### DIFF
--- a/packages/config/index.js
+++ b/packages/config/index.js
@@ -6,24 +6,23 @@ var root = require('pkg-dir').sync(process.cwd());
 
 var manifestUtil = require('./lib/manifest-util');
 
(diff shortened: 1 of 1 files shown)


ZauberNerd:
@dmbch yes, if you run your application through npm run scripts - in case that you don't run the scripts through npm it will read the package.json file and take that configuration and then it won't be overridden by config set via `npm config set ...` - although maybe you're right and this isn't needed.

dmbch:
@ZauberNerd do you still think this is needed? Otherwise, would you mind closing..? 💚

PR #224, opened 2017-09-19
Release 6.2.6

dmbch:


dmbch:
@ZauberNerd @robin-drexler PTAL

robin-drexler:
I assume the reason to do the update now is to get a fix for https://github.com/webpack/webpack-dev-server/issues/1084, am I right?

The commit message could reflect that. :)

Otherwise, lgtm.

Issue #223, opened 2017-09-15
hops@next Ideas on how to separate concerns in hops packages


ZauberNerd:
hops@next

1. global-cli (npm package name: hops)
  - globally installed
  - runs update check in background to notify about new version of itself
  - contains just the bootstrapping part of `hops init` that when executed
    installs all other dependencies and executes init command from local cli
    (similar to [react-native-cli](https://github.com/facebook/react-native/blob/master/react-native-cli/index.js) / [create-react-app](https://github.com/facebookincubator/create-react-app/blob/master/packages/create-react-app/createReactApp.js)).
  - commands:
    - `hops init` - create a new project or add hops to an existing project
      - creates a sample hops project - template based - maybe uses slim
        version of "demo" as base template
      - can also update "package.json" file in existing project by adding
        dependencies, config, scripts etc

2. local-cli
  - contains code to initialize new projects
  - commands that are being loaded into global cli:
    - `hops build` - creates the JS browser bundles and/or app shells
    - `hops develop` - runs a local webpack dev server
    - `hops serve` - starts an express server (to be used in production)

3. hops-build
  - contains build/develop webpack config
  - contains webpack dev server
  - should be installed as a devDependency
  - should be the only package that relies on webpack, babel, etc

4. hops-server
  - contains `rewritePath` / `registerMiddleware` and other shared functionality

5. hops-express
  - contains express server
  - exposes express's `app` (which can be used for aws-serverless-express)
  - exposes JS entry file to start the express server
  - can be an optional dependency if no server is needed

6. hops-lambda
  - contains cloudformation stack / lambda config (?)

4. hops-template-default / hops-template-*
  - fully functional demo/example projects that can be used to init a new project

- A typical hops project would therefore have the following dependencies:
  - `hops-react`
  - `hops-redux`
  - `hops-common` (4.)
  - `hops-build` (3.) - (as a dev dependency) contains webpack config / dev server
  - `hops-server` (2.) - (if needed) contains node server
  - `react(-dom|helmet|redux|router|router-dom)`, `redux(-thunk)`

ZauberNerd:
Work on this has started in: https://github.com/ZauberNerd/hops/commits/hops-split

dmbch:
#225 is merged 🌈

PR #225, opened 2017-09-20
[WIP]: DO NOT MERGE - split out server/build/cli modules

ZauberNerd:
This is based on #223


packages/cli/index.js
```diff
@@ -1,16 +1,29 @@
 'use strict';
 
+var hopsBuild = require('hops-build');
```
dmbch: Can't we keep this in another file? I'd love to not include that express stuff if we don't actually need it.

PR #226, opened 2017-09-22
feat(hops-react): Improve template data injection

fuko-fabio:
Small change to hops-react package regarding template data escaping

Currently for all injected globals we are escaping `<` sign.
According to doc I found it's better to escape `</`

> To prevent against this hole, you should replace every occurrence of </ in your JSON with <\/ so that the <script> tag remains open. (The characters < and / are valid only within a string literal so the replacement can’t affect anything else.)

This also allows me to inject data on rails based backend. Example:
```js
const ctx = Context.extend({
  getTemplateData () {
    const templateData = Context.prototype.getTemplateData.call(this);
    return Object.assign(templateData, {
      globals: templateData.globals.concat([{
        name: 'MY_NAME',
        value: '<%= @my_rails_data.html_safe %>'
      }])
    });
  }
});
```

### DIFF
--- a/packages/react/lib/template.js
+++ b/packages/react/lib/template.js
@@ -1,7 +1,7 @@
 'use strict';
 
 function esc (data) {
-  return JSON.stringify(data).replace(/</g, '\\u003c');
+  return JSON.stringify(data).replace(/<\//g, '\\u003c');
 }
 
 module.exports = function (data) {


dmbch:
Hi @fuko-fabio, thanks for your PR! I'm afraid I can't accept it - as far as I can tell, your change will only work for a subset of the cases we're targeting now.

FYI: we will switch to a "proper" solution for escaping with the next release, specifically this one..:
https://www.npmjs.com/package/serialize-javascript

With regards to your Rails use case: could you just change template delimiters (`<%=`) to something HTML safe?

fuko-fabio:
@dmbch can you tell me when you are planning next release with `serialize-javascript`?

> With regards to your Rails use case: could you just change template delimiters (<%=) to something HTML safe?

For this I found another solution so it's fine :)

PR #227, opened 2017-09-22
[WIP] fix(demo): add missing babel-core dependency needed by babel-jest

robin-drexler:
```
npm WARN babel-jest@21.0.2 requires a peer of babel-core@^6.0.0 || ^7.0.0-alpha || ^7.0.0-beta || ^7.0.0 but none is installed. You must install peer dependencies yourself.
```
`babel-jest` itself is a dependency of `jest-preset-hops`.

PR #228, opened 2017-09-25
fix(jest-preset-hops): add babel-core to satisfy peer dependencies

robin-drexler:
In favor of #227 

Thanks @dmbch :)


robin-drexler:
It appears, there is no issue with npm not updating package-lock.json on install.

The file is gitignored. 

```
 hops git:(add-babel-core-jest) git check-ignore -v -- demo/package-lock.json
.gitignore:5:package-lock.json	demo/package-lock.json
```

PR #242, opened 2017-10-17
feat(build-config): added mode as environment variable

matthias-reis:
## Current state

NODE_ENV is accepted as a global variable and therefore is used for code optimizations in the minifying step. But we also need it for a global variable called HOPS_MODE, which indicates whether the application runs in a `static` build environment or a `dynamic` middleware use case.

## Changes introduced here

Added the environment variable `HOPS_MODE` with a default of `dynamic` to all webpack configs.

## Checklist

### DIFF
--- a/packages/local-cli/commands/develop.js
+++ b/packages/local-cli/commands/develop.js
@@ -8,6 +8,12 @@ module.exports = function developCommand (callback) {
     describe: 'Starts a webpack-dev-server to enable local development with ' +
       'hot code reloading',
     builder: {
+      static: {
+        alias: 's',
+        default: false,
+        describe: 'Serve app in hot mode with static env variable turned on',
+        type: 'boolean'
+      },
       clean: {
         alias: 'c',
         default: true,
@@ -17,6 +23,8 @@ module.exports = function developCommand (callback) {
       }
     },
     handler: function developHandler (argv) {
+      process.env.HOPS_MODE = argv.static ? 'static' : 'dynamic';
+
       hopsBuild.runServer(argv, callback);
     }
   };
--- a/packages/local-cli/commands/serve.js
+++ b/packages/local-cli/commands/serve.js
@@ -9,6 +9,12 @@ module.exports = function serveCommand (callback) {
     'application',
     builder: {
       production: {
+        static: {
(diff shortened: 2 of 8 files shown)


packages/local-cli/commands/serve.js
```diff
@@ -20,6 +20,8 @@ module.exports = function serveCommand (callback) {
       if (argv.production) {
         process.env.NODE_ENV = 'production';
       }
+      // production environment always refers to starting dynamic mode
```
dmbch: I'm afraid that's not really true - Hops does, in fact support, static production mode. For that case, we even have some specific code in place to handle expires headers for pre-generated HTML files: https://github.com/xing/hops/blob/master/packages/express/app.js#L20

packages/local-cli/commands/start.js
```diff
@@ -8,6 +8,12 @@ module.exports = function startCommand (callback) {
     command: 'start',
     describe: 'Starts a development or production server, based on NODE_ENV',
     builder: {
+      static: {
+        alias: 's',
+        default: false,
+        describe: 'Sets mode to static if production is not turned on',
```
dmbch: See above :)

packages/local-cli/commands/start.js
```diff
@@ -28,8 +34,11 @@ module.exports = function startCommand (callback) {
         process.env.NODE_ENV = 'production';
       }
       if (process.env.NODE_ENV === 'production') {
+        // production environment always refers to starting dynamic mode
+        process.env.HOPS_MODE = 'dynamic';
```
dmbch: See above :)

packages/local-cli/commands/develop.js
```diff
@@ -16,7 +16,9 @@ module.exports = function developCommand (callback) {
         type: 'boolean'
       }
     },
-    handler: function developHandler (argv) {
+    handler: function developHandler(argv) {
```
dmbch: ESLint chokes on this line: `Missing space before function parentheses`

matthias-reis:
@dmbch PTAL

dmbch:
@matthias-reis would you mind rebasing..? :)

PR #252, opened 2017-10-26
chore: specify npm registry in lerna config

ZauberNerd:
Because of a bug in lerna the registry environment variable will be set
to yarnpkg in certain cases, therefore we specify the registry manually
in the lerna config.

Related: https://github.com/lerna/lerna/issues/896


dmbch:
Thanks! 💚

PR #254, opened 2017-10-26
Fix render context life cycle

dmbch:
## Current state

The last hops update changed the behaviour of the main render method(s).

## Changes introduced here

This PR restores the previous behaviour, while introducing another render context life cycle method, `prepareEnhancedElement`.

## Checklist

### DIFF
--- a/packages/graphql/node.js
+++ b/packages/graphql/node.js
@@ -8,8 +8,8 @@ var common = require('./lib/common');
 var fragmentsFile = require('./lib/util').getFragmentsFile();
 
 exports.Context = exports.createContext = common.Context.extend({
-  prepareEnhancedElement: function (enhancedComponent) {
-    return ReactApollo.getDataFromTree(enhancedComponent);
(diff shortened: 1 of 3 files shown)


packages/graphql/node.js
```diff
@@ -8,7 +8,7 @@ var common = require('./lib/common');
 var fragmentsFile = require('./lib/util').getFragmentsFile();
 
 exports.Context = exports.createContext = common.Context.extend({
-  bootstrap: function (enhancedComponent) {
+  prepareEnhancedElement: function (enhancedComponent) {
```
robin-drexler: Is `prepareEnhancedElement` really the way we want to advertise this method? The name invites to mutate the `enhancedElement` whereas it's (currently) used to do derive some data out of it.

How about something like `beforeRender` or `willRender`?

robin-drexler: Or return the mutated object (or newly created) and use it to render.

PR #256, opened 2017-10-26
Dependency Updates

dmbch:
## Current state

Some Hops dependencies are out of date - the most obvious of them being React. React 16 has been released for a couple of weeks now and various experiments to update Hops project have been unproblematic.

## Changes introduced here

All Hops dependencies have been updated to their latest versions. React 15, however, remains fully supported: hops-react and hops-redux peer-dependencies have not been bumped. Rather, a version range has been added for React packages: `>=15.0.0 <17.0.0`

## Checklist


packages/graphql/package.json
```diff
@@ -28,6 +28,7 @@
   },
   "peerDependencies": {
     "graphql-tag": "^2.5.0",
+    "react": ">=15.6.2 <17.0.0",
```
ZauberNerd: Why not `>=15.0.0`? This now seems to imply that something changed in `15.6.2` and we can't support previous versions.

packages/react/dom.js
```diff
@@ -44,7 +44,11 @@ exports.render = function (reactElement, context) {
     return context.bootstrap().then(function () {
       var enhancedElement = context.enhanceElement(reactElement);
       return context.prepareRender(enhancedElement).then(function () {
-        ReactDOM.render(enhancedElement, mountpoint);
+        if (ReactDOM.hydrate) {
```
ZauberNerd: Is it ok to call `.hydrate()` if there's no server-side rendered markup?
I assume it would just work because then there's a tree mismatch at the root level and everything will have to be rerendered. Does it print any warnings or is there maybe a performance penalty for doing this instead of using `.render()` in a purely client-side application?

yarn.lock
```diff
@@ -7556,7 +7574,7 @@ yargs@^8.0.2:
     y18n "^3.2.1"
     yargs-parser "^7.0.0"
 
-yargs@^9.0.0, yargs@^9.0.1:
+yargs@^9.0.0:
```
ZauberNerd: Why do we still have `yargs@9`? Did you delete and recreate yarn.lock after updating the package.json files?

packages/react/dom.js
dmbch: I think we can always expect *some* sort of server side rendering to have occurred in a Hops context - either at build- or runtime. Accordingly, we should be safe to always `hydrate`, shouldn't we..?

yarn.lock
dmbch: I did - turns out we have a couple of yargs in our dependency tree..:

```
$ yarn list yargs
├─ flow-typed@2.2.1
│  └─ yargs@4.8.1
├─ jest-cli@21.2.1
│  └─ yargs@9.0.1
├─ jest-runtime@21.2.1
│  └─ yargs@9.0.1
├─ lerna@2.4.0
│  └─ yargs@8.0.2
├─ uglify-js@2.8.29
│  └─ yargs@3.10.0
├─ webpack-dev-server@2.9.3
│  └─ yargs@6.6.0
├─ webpack@3.8.1
│  └─ yargs@8.0.2
└─ yargs@10.0.3
```

packages/react/dom.js
dmbch: Btw. I just verified that hydrating without pre-rendered markup apparently works, but according to the [ReactDOM docs](https://reactjs.org/docs/react-dom.html#hydrate), "you should treat mismatches as bugs and fix them".

Once again, in a hops-react context, I would always expect server side rendering to occur. The above-mentioned docs contain some hints on how to deal with necessary differences in server/browser rendering.

PR #257, opened 2017-10-26
fix(graphql): add lib folder to package

dmbch:
## Current state

hops-graphql is currently completely broken because its `lib` folder is missing.

## Changes introduced here

`lib` folder is included in npm package.

## Checklist

### DIFF
--- a/packages/graphql/package.json
+++ b/packages/graphql/package.json
@@ -16,7 +16,9 @@
   "browser": "dom.js",
   "files": [
     "node.js",
-    "dom.js"
+    "dom.js",
+    "commands",
+    "lib"
   ],
   "dependencies": {
     "apollo-cache-inmemory": "^1.0.0",


packages/graphql/package.json
```diff
@@ -16,7 +16,8 @@
   "browser": "dom.js",
   "files": [
     "node.js",
-    "dom.js"
+    "dom.js",
+    "lib"
```
ZauberNerd: Should we maybe switch to using `.npmignore` files instead to not run into the same issue again?

packages/graphql/package.json
```diff
@@ -16,7 +16,8 @@
   "browser": "dom.js",
   "files": [
     "node.js",
-    "dom.js"
+    "dom.js",
```
ZauberNerd: Also: The `commands/` folder is missing in `"files"`?

dmbch: I should probably simply got to bed... 😴

packages/graphql/package.json
dmbch: Sounds good, but I wouldn't want to do that just now.

dmbch:
@ZauberNerd could you take another look? 🤓

Issue #260, opened 2017-10-30
Correct Lifecycle flow at bootstrap time


matthias-reis:
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

Before Hops V 7, there was a clear order for the calls of lifecycle methods. The async `bootstrap` had been called _and finished_ before `enhanceElement`. This is currently used to fetch async data before rendering to the DOM, so Hops 7 it breaks in current applications.

## Proposed solution

Three scenarios should be supported:

- doing asynchronous stuff __before creating__  the element (see example above)
- doing asynchronous staff __before rendering__ and __after creating__ the element (usecase: apollo graphql client)
- composition must still be supported

__Possible solution:__
Turn `enhanceElement` into a Promise, that receives a Promise from the chain before and returns again a Promise. This Promise should in any case resolve to a renderable Element.

__Example (no runnable code, just to explain the principle):__

```jsx
// using jsx for clarification, should be es5
enhanceElement: function (prev) {
  let data = {};

  return fetchSth()
    .then(remoteData => data = remoteData)
    .then(prev)
    .then(el => <MyProvider data={data}>{el}</MyProvider>)
    .then(el => {
      return doSthAsync()
        .then(Promise.resolve(el));
    });
}
```

dmbch:
Hey @matthias-reis, thanks for reporting this! I really like your idea, but I thought this was already fixed in https://github.com/xing/hops/pull/254 and released with hops-react v7.3.2. Can you confirm?

matthias-reis:
Yepp, you"re right ...

PR #265, opened 2017-11-06
Do Not Merge: Mixinable Context

dmbch:
## Current state

Currently, hops only supports simple inheritance: hops-graphql inherits from hops-redux inherits from hops-react. There is no way to remove Redux from the chain. This is exacerbated for projects that use custom contexts, e.g. to handle translations.

## Changes introduced here

With this PR, we start moving to a compositive context inheritance model based on mixins. The basic React-context can now be enhanced with different mixins (e.g. redux, graphql, intl) while being able to control mixin composition strategies.

## Checklist


dmbch:
@matthias-reis, @robin-drexler, @ZauberNerd what do you think?

In my opinion, making the context extensible using this mixin approach gives us a rather nice, straightforward way to compose different kinds of functionality. On the other hand, a bit more magic and hidden complexity...

packages/react/lib/common.js
```diff
@@ -1,25 +1,14 @@
...
-    prototype: Object.assign(
-      Object.create(parent.prototype),
-      protoExtension,
-      {
-        constructor: Context
-      }
-    )
-  });
-};
+exports.Context = mixin({
```
matthias-reis: The helper methods in your package are very useful, I think (although I'm sure, that there are packages outside, that can do the same thing - or native code, that is not much longer). Mixin itself on the other hand seems to be another class implementation, which landed in node at version 4.0.0 (http://node.green/#ES2015-functions-class).

dmbch: `class` is missing in IE, though, so we will not be able to use it in hops anytime soon. And I am still conviced that limiting ourselves to just one, pseudo-classical, form of inheritance in JS is not a good idea. #klassenkampf

What 3rd party implementations of those helpers are you thinking of? I would love to keep mixinable as small as possible...

matthias-reis: Oh ... then I have a misunderstanding of the flow. Isn't this transpiled by webpack for the browser?

And I'm not completely sure what you mean by 'And I am still conviced that limiting ourselves to just one, pseudo-classical, form of inheritance in JS is not a good idea.'... Should we make the API complicated to learn by intention?

dmbch: No, the hops stuff is not transpiled. What I am replacing with this this PR is the previous Backbone-style `extends` implementation. 

And no we should not complicate things arbitrarily, of course. But having `class`es in ES6+ does, in my opinion, not imply we have to stop using the actual underlying inheritance mechanism, i.e. prototypal inheritance. `class`es are "just" syntactic sugar on top of that tried and tested mechanism, after all.

In fact, I think that with this mechanism, we might actually simplify things for our users. This is what one would have to do to compose different contexts after this PR..:

```javascript
import { createContext } from 'hops-react';
import reduxMixin from 'hops-redux';
import graphqlMixin from 'hops-graphql';

export default createContext.mixin(
  reduxMixin,
  graphqlMixin
);
```

To achieve the same compositional pattern with `class`es, we would have to do things like..:

```javascript
import { Context as ReactContext } from 'hops-react';
import { Context as ReduxContext } from 'hops-redux';
import { Context as GraphQLContext } from 'hops-graphql';

export default class Context extends ReactContext {
  enhanceElement(reactElement) {
    return super.enhanceElement()
      .then(ReduxContext.prototype.enhanceElement.bind(this))
      .then(GraphQLContext.prototype.enhanceElement.bind(this));
  }
  getTemplateData() {
    return {
      ...super.getTemplateData(),
      ...ReduxContext.prototype.getTemplateData.call(this)
      ...GraphQLContext.prototype.getTemplateData.call(this)
    };
  }
}
```

Developing additional mixins is, in my opinion, rather similar to what one would do using `class`es. Am I making some sort of sense?

matthias-reis: I know what you want to accomplish, but the problem is, that `react` obviously always has to be the root in order to get this to work (as all packages currently depend on react and it is currently the only 'Mixinable' if I see correctly). If some other mixin depends on any other (like e.g. router on redux or dataEnhancer on redux), you'll get to the babel problem (every module is a dedicated plugin, but they all have to be **applied in the correct order**).

Second point: By using mixin, you can enhance an object however you want, e.g. with **additional non lifecycle functions**. Let's say mixin `fooDataEnhancer` implements a function `createEnhancer`. When using this together with hops-redux, it will (correct me if I'm wrong) not be added to the object and therefore not work if it's added after hops redux or kill hops redux when added before.

And also: Each of the lifecycle methods **can and does alter local state** of the context instance. So you can load something in `bootstrap` and later on rely on it in `enhanceElement`. I think it's probably better to be pure and uniform in a way, e.g. like making all lifecycle methods equally async and make them return data that can later on be picked up by the next lifecycle. 

__So isn't it better to go one step back and go simpler with plain objects?__

The object would contain keys of lifecycle hooks - super easily testable (even outside a lifecycle) as they only get input and deliver output with `then`.

The usage for an app would still be easy

```jsx
const myContext = [
  reactContext, // ?
  // reduxContext,
  graphQlContext,
  {
    bootstrap: function () {
      //do something funny
    }
    enhanceElement(el) {
      return Promise.resolve(<Enhancer>{el}</Enhancer>);
    }
  }
]

render(<MyRoot />, myContext)
```

Because an intelligent, central manager coordinates everything like
- stitching all together (i.e goes through all contexts and executes their bootstrap then goes to the next lifecycle)
- using the right method (parallel, serial or merge) to combine the promises
- handing over data to the next lifecycle hook if necessary
- calling stuff differently on the client and on the server
- ...

dmbch: Good points from where I'm standing - I'll take that approach for a spin in the next couple of days (or rather nights).

dmbch:
@matthias-reis, @robin-drexler, @ZauberNerd feel free to take a peek whenever you have the time.

Some of the more interesting places might be:
* ["abstract" mixin definition](https://github.com/xing/hops/blob/398e46b1f5a619127e3e8a2014d0c195aebd58b8/packages/react/node.js#L13), where mixin application strategies are being supplied. Please note that there are seperate defintions for node and browser, but that only applies to `hops-react`
* [context creator extension](https://github.com/xing/hops/blob/398e46b1f5a619127e3e8a2014d0c195aebd58b8/packages/graphql/node.js#L49), where `hops-graphql` features are being mixed into the default
* [context mixin composition](https://github.com/xing/hops/blob/398e46b1f5a619127e3e8a2014d0c195aebd58b8/packages/template-react/src/context.js), where both `hops-redux` and `hops-graphql` mixins are applied

Regarding that last example: one could obviously also just extend the `createContext` export of either `hops-redux` or `hops-graphql` with the respective other `mixin` export. Sort of like the eternal question: *Bacon One with Cheese* or *Cheesy One with Bacon*? 🍔

dmbch:
Btw, @matthias-reis, almost all exposed lifecycle methods have now somehow become asynchronous, something you requested for `enhanceElement` in #260. The exception is `getMountpoint`, but making that one async would not make much sense in any szenario I can think of.

dmbch:
@matthias-reis, @ZauberNerd: I renamed a couple of exports and adapted `template-react` following todays discussion. PTAL and let me know what you think :)

packages/template-react/src/context.js
```diff
@@ -0,0 +1,19 @@
+// @flow
```
ZauberNerd: I'd like to go without this file for now, especially, since we agreed, to remove graphql from the `template-react` package anyways, therefore, in future, we could just `import { createContext } from 'hops-redux';`. Could you inline this file in the `app.js`?

ZauberNerd: Or maybe we can prepare a separate PR, that removes graphql from the default template and merge that PR, before we merge this.

ZauberNerd:
@dmbch currently `hops-graphql` still contains `hops-redux` as a dependency, because it extends the `hops-redux` context - please remove that as well.

packages/react/dom.js
```diff
@@ -3,37 +3,43 @@
...
 
-exports.Context = exports.createContext = Context.extend({
-  initialize: function (options) {
+exports.contextDefinition = {
+  constructor: function (options) {
+    if (!options) { options = {}; }
     this.mountpoint = options.mountpoint || '#main';
   },
   bootstrap: function () {
     return Promise.resolve();
```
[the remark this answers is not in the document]
dmbch: If we want to force a mixinable method to return a Promise, we unfortunately have to have these no-ops. Other than that, the default mixinable strategies handle that case.

packages/template-react/src/context.js
dmbch: I think you are currently splitting the templates yourself, so yeah, let's do that,

Issue #266, opened 2017-11-07
Give devServer config higher priority


ghost23:
## This is my intent (choose one)

- [ ] I want to report a bug
- [x] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

In the server.js:

https://github.com/xing/hops/blob/v7.4.0/packages/build/server.js#L22

currently, after the devServer config options there is an additional object (the one with the after function and the watchOptions). This way, it is impossible to override that functionality. For example, you are always stuck with the rewrite:

https://github.com/xing/hops/blob/v7.4.0/packages/build/server.js#L24

even if you don't want to use that.

## Proposed solution

I would like to propose to simply change the order here:

https://github.com/xing/hops/blob/v7.4.0/packages/build/server.js#L22

from:
```Object.assign({}, config.devServer, { after: .... })```
to:
```Object.assign({}, { after: .... }, config.devServer)```

It wouldn't affect existing setups (nobody currently could have provided his/her own after: function, it would have been overridden)

(cc: @mjacobus)

dmbch:
Sounds good! Would you like to provide a PR yourself?

ghost23:
Yes, I will.

dmbch:
Cool, thanks! 🚀

Please remember to add your info to the contributors section in the main `package.json` file 💡

ghost23:
Done

dmbch:
Thanks! 

Fixed in https://github.com/xing/hops/blob/v7.4.1/packages/build/server.js#L22
Released as hops-build v7.4.1: https://www.npmjs.com/package/hops-build

PR #267, opened 2017-11-08
Give devServer config higher priority

ghost23:
This pull request closes issue #266 

## Current state

In the server.js:

https://github.com/xing/hops/blob/v7.4.0/packages/build/server.js#L22

currently, after the devServer config options there is an additional object (the one with the after function and the watchOptions). This way, it is impossible to override that functionality. For example, you are always stuck with the rewrite:

https://github.com/xing/hops/blob/v7.4.0/packages/build/server.js#L24

even if you don't want to use that.

## Changes introduced here

This change simply changes the order here:

https://github.com/xing/hops/blob/v7.4.0/packages/build/server.js#L22

from:
Object.assign({}, config.devServer, { after: .... })
to:
Object.assign({}, { after: .... }, config.devServer)

So now the devServer configs can override even the 'after' function and the watchOptions, if it so wishes.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
     "BlakeSimpson <blakersim@gmail.com> (https://www.xing.com/profile/Blake_Simpson)",
     "systemboogie <marcus@marcusnoll.de> (https://www.xing.com/profile/Marcus_Noll3)",
     "KingHenne <mail@hendrik-liebau.de> (https://www.xing.com/profile/Hendrik_Liebau2)",
-    "ZauberNerd <zaubernerd@zaubernerd.de> (https://www.xing.com/profile/Bjoern_Brauer5)"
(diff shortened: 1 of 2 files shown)


dmbch:
Hi @ghost23, thanks for your PR! 🌈 

Could you please `git rebase --interactive` your commits to make sure your commit messages adhere to the [convertional format](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit)?

They ought to be similar to these..:
`refactor(build): change order of applied server config objects`
`chore: add @ghost23 to contributors`

ghost23:
Done

ghost23:
By the way, the link you gave above, doesn't point to anything meaningful. You should update it to:

https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#commits

PR #269, opened 2017-11-08
Improved renderer

dmbch:
## Current state

Static renderer does not apply bootstrapServer function from hops-config. In this regard, its implementation differs from our servers.

## Changes introduced here

Renderer now applies the `bootstrapServer` function, aligning it with server behaviour. It does so by actually instantiating an [Express Router](http://expressjs.com/en/4x/api.html#router) and using it for rendering.

N.B.: I deliberately decided to not also apply `teardownServer`, because the primary use-case of that function is to set up error handling and logging. We are doing both ourselves in a static rendering context.

## Checklist

### DIFF
--- a/packages/renderer/index.js
+++ b/packages/renderer/index.js
@@ -2,12 +2,23 @@
 
 var events = require('events');
 
+var express = require('express');
 var mocks = require('node-mocks-http');
 
 var createMiddleware = require('hops-middleware');
 
-module.exports = function createRenderer (webpackConfig, watchOptions) {
-  var middleware = createMiddleware(webpackConfig, watchOptions);
+module.exports = function createRenderer (options) {
+  var hopsConfig = options.hopsConfig || {};
+  var router = new express.Router();
+
+  if (hopsConfig.bootstrapServer) {
+    hopsConfig.bootstrapServer(router, hopsConfig);
+  }
+
+  router.use(createMiddleware(
+    options.webpackConfig,
+    options.watchOptions
+  ));
 
   return function (location) {
     return new Promise(function (resolve, reject) {
@@ -25,7 +36,7 @@ module.exports = function createRenderer (webpackConfig, watchOptions) {
           resolve(res._getData());
         }
       });
-      middleware(req, res, function (error) {
+      router.handle(req, res, function (error) {
         if (error) { reject(error); } else { resolve(); }
       });
     });
--- a/packages/renderer/index.js
+++ b/packages/renderer/index.js
@@ -20,11 +20,12 @@ module.exports = function createRenderer (options) {
     options.watchOptions
   ));
 
-  return function (location) {
+  return function (options) {
+    if (typeof options === 'string') {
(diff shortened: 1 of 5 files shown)


packages/renderer/index.js
```diff
@@ -2,12 +2,23 @@
...
 
+var express = require('express');
 var mocks = require('node-mocks-http');
 
 var createMiddleware = require('hops-middleware');
 
-module.exports = function createRenderer (webpackConfig, watchOptions) {
-  var middleware = createMiddleware(webpackConfig, watchOptions);
+module.exports = function createRenderer (options) {
+  var hopsConfig = options.hopsConfig || {};
```
robin-drexler: If `hopsConfig` was undefined or falsey, it would already break in https://github.com/xing/hops/pull/269/files#diff-1951c43c9dcbe1f7c6ae7bbb666ef319R24 wouldn't it?

dmbch: Wait, why?

dmbch: If options was falsey it would break, but that is fine for me.

packages/spec/renderer.js
```diff
@@ -16,19 +16,19 @@ describe('renderer', function () {
   });
 
   it('should create a renderer function', function () {
```
robin-drexler: Do you think a test for the newly introduced `bootstrapServer` feature could easily be added?

packages/renderer/index.js
```diff
@@ -2,12 +2,23 @@
...
+  if (hopsConfig.bootstrapServer) {
+    hopsConfig.bootstrapServer(router, hopsConfig);
+  }
+
+  router.use(createMiddleware(
+    options.webpackConfig,
+    options.watchOptions
+  ));
 
   return function (location) {
```
robin-drexler: I think we need to be able to pass more than location here to configure the request.
For example, in order to be able to change the language, we need to set headers (cookie or accept-language)

packages/spec/renderer.js
```diff
@@ -16,19 +16,35 @@ describe('renderer', function () {
...
 
+    it('should call middleware', function () {
+      var render = createRenderer({
+        webpackConfig: goodConfig,
+        hopsConfig: {
+          bootstrapServer: function (app) {
+            assert(true);
+            app.use(function (req, res, next) {
+              assert(true);
+              next();
```
robin-drexler: The test is green even if `bootstrapServer` is never called, because the test is synchronously.
How about adding a `done()` callback here, so the test will timeout when the middleware is never called?

packages/spec/renderer.js
```diff
@@ -36,6 +36,7 @@ describe('renderer', function () {
             app.use(function (req, res, next) {
               assert(true);
               next();
+              done();
```
dmbch: @robin-drexler nice catch - should be good now, shouldn't it?

robin-drexler: Indeed! :)

Issue #277, opened 2017-11-22
Remove flow from hops-template-react


ZauberNerd:
## This is my intent (choose one)

- [x] I want to provide or change a feature

## The problem

The current `hops-template-react` contains flow type annotations.

## Proposed solution

Remove flow support and flow types from current template `hops-template-react`.

Issue #276, opened 2017-11-22
Remove graphql from hops-template-react


ZauberNerd:
## This is my intent (choose one)

- [x] I want to provide or change a feature

## The problem

The current `hops-template-react` template contains a lot more than just react and redux.

We should remove graphql from the `hops-template-react` and create a new template `hops-template-graphql` that only contains react and graphql.

**Question:** Do we want the new template `hops-template-graphql` to contain:
- [x] just react and graphql
- [ ] react, redux and graphql

## Proposed solution

Split up `hops-template-react` into multiple templates.

PR #279, opened 2017-11-22
Simplify hops-template-react

ZauberNerd:
This PR removes graphql and flow type annotations from the `hops-template-react`.

Related: #276 #277


ZauberNerd:
@dmbch done - please take a look.

PR #283, opened 2017-11-24
Improve Readmes

dmbch:
## Current state

Currently, our READMEs are, in places, incomplete and outdated.

## Changes introduced here

In this PR, we want to start improving the READMEs in these regards.

## Checklist

- [ ] All code is written in plain ECMAScript v5 and adheres to the [semistandard format](https://github.com/Flet/semistandard)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ Hops is quite modular, its component being independently published on npm. Its m
 | [Hops Redux](https://github.com/xing/hops/tree/master/packages/redux)               | [![npm](https://img.shields.io/npm/v/hops-redux.svg)](https://www.npmjs.com/package/hops-redux)               |
 | [Hops GraphQL](https://github.com/xing/hops/tree/master/packages/graphql)           | [![npm](https://img.shields.io/npm/v/hops-graphql.svg)](https://www.npmjs.com/package/hops-graphql)           |
 | [Hops Express](https://github.com/xing/hops/tree/master/packages/express)           | [![npm](https://img.shields.io/npm/v/hops-express.svg)](https://www.npmjs.com/package/hops-express)           |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -34,74 +34,33 @@ This will start hops in development mode. Visit [http://localhost:8080](http://l
...
-```shell
-npm start (--production)
-```
-
-If called with the `--production` flag, a production mode server is launched. Otherwise, a development server featuring hot module replacement is started. Hops can generate html pages for all `locations` listed in its config at build time, so you don't need a node server in production, but still can enjoy the benefits of a true universal JavaScript application.
-
-### Advanced Usage
+### General Usage
 
 Hops is quite modular, its component being independently published on npm. Its main building blocks, however, rely upon each other:
```
robin-drexler: `its component` -> `its components` (plural)

README.md
```diff
@@ -34,74 +34,33 @@ This will start hops in development mode. Visit [http://localhost:8080](http://l
...
-
-Besides reasonably recent versions of [Node.js](https://nodejs.org/en/) and [npm](https://www.npmjs.com), hops has no global dependencies. If you need those, we recommend using [nvm](https://github.com/creationix/nvm) or similar.
-
-Install the `hops-cli` globally via npm `npm install --global hops-cli` or yarn `yarn global add hops-cli`:
-
-### Project creation
-
-`hops init my-project` will create a directory called `my-project` and initialize an example react project inside it.
-
-### Configuration
```
robin-drexler: I kinda liked the config intro, because it's an integral part a hops app. Does anything speak against keeping it?

dmbch: I was thinking about that, too, but I came to the conclusion that it is not more integral or important than, say, the CLI stuff.

I’d also like to extend the component tables below with short introductory texts for all of them instead of highlighting a single feature.

robin-drexler: > I’d also like to extend the component tables below with short introductory texts for all of them instead of highlighting a single feature.

<3

PR #287, opened 2017-11-28
Redux middleware option

robin-drexler:
## Current state
In order to set redux middlewares a consumer needs to create a new context that mixes with the redux context and override `getMiddlewares`.

## Changes introduced here

Since it's a pretty common use case to set middlewares, make it easier to do so by adding it as a context option. 

## Checklist

### DIFF
--- a/packages/spec/redux.js
+++ b/packages/spec/redux.js
@@ -1,48 +1,20 @@
 /* eslint-env node, mocha */
 var assert = require('assert');
-var React = require('react');
 
-var redux = require('hops-redux');
-var hopsReact = require('hops-react');
+var hopsRedux = require('hops-redux');
 
 describe('redux', function () {
-  var defaultOptions = {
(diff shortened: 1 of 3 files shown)


packages/spec/redux.js
```diff
@@ -1,48 +1,20 @@
...
-      .enhanceElement(React.createElement('span'))
-      .then(function (element) {
-        element.props.store.dispatch({ type: '' });
-        assert.ok(called);
-      });
+    assert.equal(context.getMiddlewares()[0], middleware);
   });
 
-  it('throws an error, when middlewares is provided but not an array', function () {
+  it('throws array when middlewares is not an array', function () {
```
dmbch: I'd expect it to throw an error, don't you?

robin-drexler: Oh, indeed! Fixed.

PR #288, opened 2017-11-28
Prettier

dmbch:
## Current state

Hops does not use [prettier](https://prettier.io).

## Changes introduced here

It now does.

## Checklist

- [X] All code is written in plain ECMAScript v5 and is formatted using [prettier](https://prettier.io).
- [X] Documentation has been added


robin-drexler:
What does "It now does." mean exactly? Where is it supposed to invoke prettier automatically?
I assumed it's now a precommit hook, but that didn't work (at first try). :D

dmbch:
@robin-drexler that is a bit weird: for me, it does work as a precommit hook. Would you mind `yarn reset`ing your clone and trying once more?

Other than that, I added a package script allowing us to `yarn fmt` the whole project.

robin-drexler:
After a fresh clone the precommit hook works fine. 👍 
However, running `yarn` resulted in a changed `yarn.lock`. Did you forget to check it in?

dmbch:
@robin-drexler no, but I think I `prettier`ed it, too...

PR #290, opened 2017-11-30
Context cloning

dmbch:
## Current state

Currently, cloned context mixins have to take care of merging options.

## Changes introduced here

With the proposed change, context mixin constructors receive a single merged options hash.

## Checklist

### DIFF
--- a/packages/react/node.js
+++ b/packages/react/node.js
@@ -17,9 +17,10 @@ exports.combineContexts = mixinable({
   renderTemplate: mixinable.override,
 });
 
-exports.ReactContext = function() {
-  var args = Array.prototype.slice.call(arguments);
-  var options = Object.assign.apply(Object, [{}].concat(args));
+exports.ReactContext = function(options) {
+  if (!options) {
+    options = {};
+  }
   this.template = options.template || defaultTemplate;
   this.request = options.request;
   this.routerContext = {};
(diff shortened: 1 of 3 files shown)


robin-drexler:
That'd be a breaking change, wouldn't it?

dmbch:
@robin-drexler technically yes, realistically probably not, as it is only altering an undocumented and thus private-ish API.

dmbch:
@KingHenne what do you think? Important improvement or just cosmetics?

KingHenne:
@dmbch To be honest I don't quite understand how `mixinable.replicate` works. But it seems this change means the options are passed to the context already merged so I don't have to merge them in my constructor? In this case I think it's a valuable improvement and not only "cosmetics".

KingHenne:
Alright, after reading the [test](https://github.com/dmbch/mixinable/blob/41c7e8dcea3613dc38bb8eb8f6e4155b9b577a53/test.js#L469-L488) I get it now. Looks good then! ... Although I find this meta-programming with `mixinable` quite hard to understand. 😕

dmbch:
@KingHenne thanks for your input! 

To be honest, I'd love to replace `mixininable` with something more straightforward, but I don't currently have an idea what existing lib or framework we might be using here. We need/want composition capabilities on the one hand and isolation on the other. 

Pseudo-classical inheritance (`class extends Foo {}`, `Foo.extend()`) does not really work for us, because we need to able to mix, modify and match context implementations: we need to be able to support all possible combinations of `React`, `Redux`, `GraphQL`, `I18n` and custom contexts...

So if you have a simpler idea that would be massively appreciated!

PR #297, opened 2017-12-07
Deprecate plugin, refactor

dmbch:
## Current state

Using our custom Webpack plugin, we are currently running Webpack within Webpack. That's probably not the simplest way to achieve what we want.

## Changes introduced here

Get rid of custom Webpack plugin.

## Checklist

* [X] Necessary unit tests are added in order to ensure correct behavior

### DIFF
--- a/packages/build/lib/generate.js
+++ b/packages/build/lib/generate.js
@@ -1,0 +1,52 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+
+var index = require('directory-index');
+var mkdirp = require('mkdirp');
+
+var hopsConfig = require('hops-config');
+var createRenderer = require('hops-renderer');
+
+function getFileName(location) {
+  return path.join.apply(
+    path,
+    [hopsConfig.buildDir].concat(index(location).split(path.sep))
+  );
+}
+
+function getKBs(s) {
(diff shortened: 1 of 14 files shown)


packages/build/lib/generate.js
```diff
@@ -0,0 +1,52 @@
...
+
+function getFileName(location) {
+  return path.join.apply(
+    path,
+    [hopsConfig.buildDir].concat(index(location).split(path.sep))
+  );
+}
+
+function getKBs(s) {
+  return ((encodeURI(s).split(/%..|./).length - 1) / 1024).toFixed(2);
```
ZauberNerd: Please use https://nodejs.org/api/buffer.html#buffer_class_method_buffer_bytelength_string_encoding

packages/build/lib/generate.js
```diff
@@ -0,0 +1,52 @@
...
+  return Promise.all(
+    (hopsConfig.locations || []).map(function(location) {
+      return render(location).then(function(html) {
+        if (html) {
+          return writeFile(location, html);
+        }
+      });
+    })
+  ).then(function() {
+    console.log();
```
ZauberNerd: What?

dmbch:
@ZauberNerd PTAnotherL

PR #298, opened 2017-12-08
Configurable router

dmbch:
## Current state

Currently, users have to provide their own (derived) `ReactContext` implementation if they want to pass props to ReactRouter instances.

## Changes introduced here

With this PR, we introduce a router config option allowing users to pass in ReactRouter props. Primary use-case is implementing a custom [`getUserConfirmation` function](https://reacttraining.com/react-router/web/api/BrowserRouter/getUserConfirmation-func).

## Checklist


Xiphe:
LGTM :wolf:

PR #299, opened 2017-12-11
refactor(lambda): print warning if misconfigured

ZauberNerd:


packages/lambda/lib/aws-config.js
```diff
@@ -42,15 +43,28 @@ module.exports = function getAWSConfig() {
...
+      'AWS Lambda only supports Node.js in version: ' +
+        AVAILABLE_NODE_RUNTIME.join(', ') +
+        '. Please specify one of these versions as node target in your ' +
+        'hopsConfig.node, to enable Babel to transpile to the correct version.'
+    );
+  }
+
+  if (
+    !config.domainName &&
+    !config.basePath.startsWith(config.stageName) &&
```
dmbch: Please be so kind as to ES5-ify this. Let's tackle the ES6-ification in a more coordinated fashion. ;)

PR #300, opened 2017-12-11
Dynamize ENV variables

dmbch:
## Current state

Currently, environment variables that are being used with webpack.EnvironmentPlugin are hard-coded in our Webpack configs.

## Changes introduced here

With this PR, environment variables that should be accessible in apps can simply be added to an `envVars` hash in hops-config. 

## Checklist

* [X] Necessary unit tests are added in order to ensure correct behavior


packages/config/README.md
```diff
@@ -19,26 +19,27 @@ Please find a list of the default options below. They can be set in your project
...
+| `node`          | `String`                   | [Node target version](https://www.npmjs.com/package/babel-preset-env#targetsnode). Defaults to `current`. It is used to specify the Node version which Babel should compile for                                                                                                                                                                                                                                                                                 |
+| `envVars`       | `{ HOPS_MODE: 'dynamic' }` | Environment variables meant to be passed to [`webpack.EnvironmentPlugin`](https://webpack.js.org/plugins/environment-plugin/), i.e. made accessible to your app                                                                                                                                                                                                                                                                                                 |
```
ZauberNerd: `{ HOPS_MODE: 'dynamic' }` is not a type, but the default value and therefore should be mentioned as such in the 'Description'.
For the `type` field you could do something like: `{ [key: String]: String }` or similar.

ZauberNerd: Also: Isn't `NODE_ENV` also worth mentioning in the description as default/special value?

PR #305, opened 2017-12-13
Introduce new lifecycle method: "prepareRender"

ZauberNerd:
This PR introduces a new lifecycle method to the hops-react context,
called `prepareRender`.
It is being called after the element tree has been constructed (through
the `enhanceElement` methods).


ZauberNerd:
This will be obsolete with #306

PR #306, opened 2017-12-13
Add hops-template-graphql

ZauberNerd:
This PR adds the new `hops-template-graphql` to enable quick
set-up of graphql clients.
It also changes the order in which `getTemplateData` and
`ReactDOM.renderToString` is called, to enable other implementations
to do additional tasks in `getTemplateData` which will prepare the
render tree.


ZauberNerd:
This is an alternative implementation to the proposed API change in #305

PR #314, opened 2017-12-15
refactor(cli): allow cli to be executed anywhere inside hops project

ZauberNerd:
Previously the `hops-cli` package could only be executed in the root
directory of a hops project.
With this change it will be possible to execute the hops cli from any
folder inside a hops project.

### DIFF
--- a/packages/cli/hops.js
+++ b/packages/cli/hops.js
@@ -159,11 +159,14 @@ var localCliPath = getLocalCliPath();
 
 var isInsideHopsProject = false;
 try {
-  var manifest = require(path.resolve(process.cwd(), 'package.json'));
+  var hopsRoot = require('pkg-dir').sync(process.cwd());
+  var manifest = require(path.join(hopsRoot, 'package.json'));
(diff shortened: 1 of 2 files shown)


packages/cli/hops.js
```diff
@@ -159,11 +159,14 @@ var localCliPath = getLocalCliPath();
...
 var isInsideHopsProject = false;
 try {
-  var manifest = require(path.resolve(process.cwd(), 'package.json'));
+  var hopsRoot = require('pkg-dir').sync(process.cwd());
+  var manifest = require(path.join(hopsRoot, 'package.json'));
   var dependencies = Object.keys(manifest.dependencies || {}).concat(
     Object.keys(manifest.devDependencies || {})
   );
-  isInsideHopsProject = dependencies.indexOf('hops-local-cli') > -1;
+  isInsideHopsProject = PACKAGES_TO_INSTALL.every(function(dependency) {
```
dmbch: Why, oh why, are we using a constant-ish array (PACKAGES_TO_INSTALL) with just a single element? ;)

Issue #308, opened 2017-12-14
Consolidate CLI Packages


dmbch:
## This is my intent (choose one)

- [X] I want to request a feature or change

## The problem

We currently have two cli packages - that is (at least) one too many. Additionally, both packages expose a `hops` executable - while the legacy `hops` npm package is deprecated.

## Proposed solution

We ought to consolidate local and global CLI packages in a single package that should be published as `hops`, reviving that package name. The existing CLI packages will be deprecated.

PR #315, opened 2017-12-15
(2/2) consolidate local-cli and cli into one package

ZauberNerd:
This is the second part of #308 
First #321 should be merged and published and then this PR should be merged and published.
This PR is optimistically rebased against #321.


packages/cli/package.json
```diff
@@ -1,20 +1,25 @@
...
-  "name": "hops-cli",
+  "name": "hops",
   "version": "9.1.1",
   "description": "Global CLI module to initialize new hops projects",
   "keywords": ["hops", "cli"],
   "license": "MIT",
   "bin": {
-    "hops": "hops.js"
+    "hops": "hops.js",
+    "hops-start": "commands/start.js"
```
dmbch: I think I would prefer not exposing hops-start in the global scope. What do you think?

packages/cli/hops.js
```diff
@@ -3,21 +3,25 @@
...
+
+  try {
+    return resolveCwd('hops');
+  } catch (_) {}
+
+  return null;
 };
 
-var PACKAGES_TO_INSTALL = ['hops-local-cli'];
+var PACKAGES_TO_INSTALL = ['hops'];
```
dmbch: Again, why does this need to be an array..?

packages/cli/lib/package-manager.js
```diff
@@ -17,18 +17,44 @@ function execIgnoreStdError(command, options) {
   return result.toString('utf-8').trim();
 }
 
+function isYarnAvailable() {
+  try {
+    execSync('yarn --version', { stdio: 'ignore' });
+    return true;
+  } catch (error) {
```
dmbch: Could we call that unused `error` `_`, please?

packages/cli/package.json
```diff
@@ -1,20 +1,25 @@
...
   },
+  "main": "index.js",
   "preferGlobal": true,
```
dmbch: Please remove that `preferGlobal`.

dmbch:
Oh, and could you please rebase..? :shipit:

packages/cli/lib/commands.js
```diff
@@ -16,7 +16,8 @@ module.exports = function findCommands() {
       })
       .map(function(command) {
         return path.join(binDir, command);
-      });
+      })
+      .concat(path.join(__dirname, '..', 'commands', 'start.js'));
```
dmbch: I think in this case we really ought to use `path.resolve`.

packages/build/commands/build.js
```diff
@@ -42,5 +42,10 @@ module.exports = function defineBuildCommand(args) {
 };
 
 if (require.main === module) {
-  require('hops-local-cli').run(module.exports, 'build');
+  try {
+    require.resolve('hops-local-cli');
```
dmbch: Would it not make more sense to try using `hops` first and only fall back to the legacy version if needed? Applies to all occurences.

packages/cli/lib/commands.js
dmbch: Turns out we should not. :D

PR #318, opened 2017-12-18
feat(jest-preset-hops): add jest 22 to peerDependency range

robin-drexler:
[jest 22 has been released](https://github.com/facebook/jest/blob/master/CHANGELOG.md#jest-2200). Tested jest@22 with the hops default template. `jest-preset-hops` still seems to work, so I'd like to increase the range of supported jest versions. :)

## Checklist

### DIFF
--- a/packages/jest-preset/package.json
+++ b/packages/jest-preset/package.json
@@ -11,11 +11,11 @@
   },
   "dependencies": {
     "babel-core": "^6.26.0",
-    "babel-jest": "21",
+    "babel-jest": "^22.0.4",
     "hops-build-config": "9.3.2",
     "identity-obj-proxy": "^3.0.0"
   },
   "peerDependencies": {
-    "jest": "^21.0.0 || ^22.0.0"
+    "jest": ">= 21.0.0 < 23.0.0"
   }
 }
--- a/packages/jest-preset/package.json
+++ b/packages/jest-preset/package.json
@@ -16,6 +16,6 @@
     "identity-obj-proxy": "^3.0.0"
   },
   "peerDependencies": {
-    "jest": "21"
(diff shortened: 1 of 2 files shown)


packages/jest-preset/package.json
```diff
@@ -16,6 +16,6 @@
     "identity-obj-proxy": "^3.0.0"
   },
   "peerDependencies": {
-    "jest": "21"
+    "jest": "^21.0.0 || ^22.0.0"
```
ZauberNerd: I'd prefer writing it in a way that the next major version doesn't require to add another `x.x.x || x.x.x || x.x.x`, but can just be extended, like so: `>= 21.0.0 < 23.0.0` or `21.x.x - 22.x.x`.

packages/template-graphql/package.json
```diff
@@ -6,7 +6,9 @@
   "config": {
     "hops": {
       "browsers": "last 1 Chrome versions",
-      "locations": ["/"],
+      "locations": [
```
ZauberNerd: Please execute prettier for all changed files: `yarn fmt`.

packages/template-react/package.json
```diff
@@ -6,7 +6,10 @@
   "config": {
     "hops": {
       "browsers": "last 1 Chrome versions",
-      "locations": ["/", "/counter"]
+      "locations": [
```
ZauberNerd: Please execute prettier for all files `yarn fmt`.

packages/jest-preset/package.json
```diff
@@ -16,6 +16,6 @@
     "identity-obj-proxy": "^3.0.0"
```
ZauberNerd: Shouldn't we also update `babel-jest` (in line 14)?

Issue #325, opened 2018-01-08
Shorten CSS classes for production builds


HerrBertling:
## This is my intent (choose one)

- [x] I want to provide or change a feature

## The problem

Currently, the CSS output for production builds is the same as for development environments (see https://github.com/xing/hops/blob/master/packages/build-config/sections/module-rules/postcss.js#L10 )

```js
var cssLoaderOptions = {
  importLoaders: 1,
  modules: true,
  localIdentName: '[folder]-[name]-[local]-[hash:8]',
  sourceMap: false,
};
```

Which results in quite long classnames. I'd like to obfuscate these way more to reduce the amount of characters shipped to users by `~70%` (rather wild guess, but `folder`, `name` and `local` should have around 6-8 characters like the hash in the end)

## Proposed solution

Provide another set of `cssLoaderOptions` for the production build with a different `localIdentName`, e.g. `[hash:base64:8]`.

I know this makes debugging in production a bit harder since the classnames are not that obvious. But I still think it's worth it shaving off quite some output :)

12finger:
@HerrBertling @matthias-reis 
regarding:
> .. still think it's worth it shaving off quite some output :

did you do the numbers already ?
i'd say, like out of the blue, that it wont yield a significant improvement for `prod`, as in many cases the GZIP running over it, will minimise the affect dramatically.

even Facebook, they don't seem to have a consistent approach to it 👻

looks like "Kraut & Rüben" oder: "Wie bei Hempels unterm Sofa!" ;)

![kcirtap amzok 2018-07-17 15-03-59](https://user-images.githubusercontent.com/1730022/42819150-fe619654-89d2-11e8-96d5-b0563a2d55fe.png)

dmbch:
@12finger any specific reason you are asking? At least in our setting, it is (read: will be) rather trivial to shorten class names, so why not do it? Even with GZIP, nothing beats having less to stuff to transfer.

dmbch:
Since we decided to close #327, I will close this, too.

HerrBertling:
Na gut 😆

PR #327, opened 2018-01-08
feat(build-config): use shorter classnames in production

HerrBertling:
fixes https://github.com/xing/hops/issues/325

This pull request closes issue #325

## Current state

The generated classnames are the same for all instances. This leads to quite long classnames in production.

## Changes introduced here

Shorter classnames in production. Yay 🎉 

## Checklist


dmbch:
Thanks for your PR! 🚀 

As discussed IRL, we will have to find a way to also implement this for server rendering. Maybe we really ought to simply rely on `NODE_ENV` and 'minify' the class name if it is set to `production`..?

packages/build-config/sections/module-rules/postcss.js
```diff
@@ -30,7 +30,9 @@ exports.build = {
     use: [
       {
         loader: require.resolve('css-loader'),
-        options: cssLoaderOptions,
+        options: Object.assign({}, cssLoaderOptions, {
+          localIdentName: '[hash:base64:8]',
```
dmbch: We will also have to do this for node to make sure that the generated markup is equal no matter where an in what env it is rendered.

systemboogie:
I like that 👍 
However, I want to point out that we have to properly communicate this (actually "non-breaking") change before merging because some end-to-end tests may rely on classnames.

HerrBertling:
I changed it to use the `NODE_ENV` to decide how to handle the classes :)

dmbch:
@systemboogie good point. How should we proceed? Ask @matthias-reis to include in in our next newsletter? Add a note to the README? 🤔

systemboogie:
@dmbch 
There is a proposal in Malt (1773) that has already been worked on. I would say we first have to get that through the door. After that, we can spread the news via developer/qa lists, communities of practice and in person. Then we should give the teams some time to use the new testId implementation and adjust their tests.

HerrBertling:
So we'll keep this open? Or merge it and advise people to update only if…?

systemboogie:
Maybe it is a good idea to make it an "opt-in" feature? That way we could merge and teams could decide whether to use it or not. Otherwise I would not merge at the moment.

HerrBertling:
Sounds good :)

HerrBertling:
I'll close here and redo this as a feature :)

HerrBertling:
Or not, as I talked with @ZauberNerd and @robin-drexler who'd rather keep the footprint of this low instead of adding a special opt-in feature for this setting :)

dmbch:
Repetitive, long class names counterintuitively lead to smaller gzipped downloads, therefore I am closing this for the time being.

PR #336, opened 2018-01-11
Better Config

dmbch:
## Current state

`hops-config` uses a proprietary configuration mechanism based on rather arcane features of npm/yarn.

## Changes introduced here

With this PR, we are retiring the previous config mechanism and move to [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) intead. That means that hops config can now be stored in a number of places, in a number of formats: e.g. a `.hopsrc` written in YAML or a `hops.config.js` exporting a plain JavaScript object literal.

Additionally, we introduce environment specific config support very similar to what [Babel](https://babeljs.io/docs/usage/babelrc/#env-option) are doing.

## Checklist


packages/config/README.md
```diff
@@ -48,52 +48,61 @@ All keys matching `/(config|file|dir)s?$/i` will be treated as a filesystem path
...
+{
+  "hops": {
+    "port": 3000,
+    "env": {
+      "my-environment": {
+        "port": 1337
+      }
+    }
+  }
+}
```
ZauberNerd: Should we also add a small section about extending the configuration with a baseline config and how to use placeholders?

ZauberNerd: Also: The readme of the CLI package mentions npm config - that should be rewritten as well.

packages/config/index.js
```diff
@@ -4,34 +4,68 @@ var path = require('path');
 
 var root = require('pkg-dir').sync(process.cwd());
```
ZauberNerd: Same here - `process.cwd()` is not required.

packages/config/README.md
```diff
@@ -48,52 +48,61 @@ All keys matching `/(config|file|dir)s?$/i` will be treated as a filesystem path
...
-To override the port value for your local environment without actually changing your package.json, set the respective npm environment variable...
+1. `hops` property in your package.json file
+2. `.hopsrc` (or `.hopsrc.(js|json|ya?ml)`)
+3. `hops.config.js` JS file exporting the object
 
-```
-npm config set my-application:hops:port 1337
+You can use whatever suits you best, but please only use a single location in your project's root folder. Hops does not attempt to merge configurations stored in different locations: it uses the first config it finds.
+
+## Environment Specific Configuration
```
ZauberNerd: We should also mention that this does not deep merge.

packages/config/index.js
```diff
@@ -4,34 +4,68 @@ var path = require('path');
...
+      buildDir: 'build',
+      cacheDir: 'node_modules/.cache/hops',
+    },
+    result ? result.config : {}
+  );
 }
 
-function extendConfig(config) {
-  if (npmConfig.extends) {
+function applyExtension(config) {
```
ZauberNerd: I liked the old name `extendConfig` better, because `extension` is also used for file extensions, for example and it isn't clear when just reading the name of this function, what it actually does.

packages/config/index.js
```diff
@@ -4,34 +4,68 @@ var path = require('path');
...
+    {
+      https: false,
+      host: '0.0.0.0',
+      port: 8080,
+      locations: [],
+      basePath: '',
+      assetPath: '',
+      browsers: '> 1%, last 2 versions, Firefox ESR',
+      node: 'current',
+      envVars: { HOPS_MODE: 'dynamic' },
```
matthias-reis: not using a deep merge here means, that everyone who adds env vars must ensure to configure `HOPS_MODE`, isn't it?

packages/config/index.js
```diff
@@ -4,34 +4,68 @@ var path = require('path');
...
+  var result = Object.assign({}, config);
+  if (config.extends) {
     try {
-      require.resolve(npmConfig.extends);
-      Object.assign(config, require(npmConfig.extends));
+      require.resolve(config.extends);
+      Object.assign(result, require(config.extends));
     } catch (_) {
-      Object.assign(config, require(path.join(root, npmConfig.extends)));
+      Object.assign(result, require(path.join(root, config.extends)));
```
ZauberNerd: In the past we've sometimes run into problems here, mainly when the `hops-config` is installed in the wrong folder or something similar.
Would it make sense to `try/catch` the second require too and print a friendly warning?

packages/config/index.js
matthias-reis: As it is a breaking change anyway, we could think of getting rid of this mechanics anyway and propagate only one way of configuration (user can choose between js, yaml or package.json anyway).

Other possibility: extend this to an eslint like structure and allow several additional configs as an array. That way, a user config (like `~/hopslocal.js`) would be possible again.

packages/config/index.js
```diff
@@ -40,14 +74,13 @@ function resolvePaths(config) {
         }
       })(config[key]);
     });
-  return config;
+  return result;
 }
 
 function normalizeURLs(config) {
   var basePath = config.basePath.replace(/^\/*/, '/').replace(/\/*$/, '');
```
ZauberNerd: Could we change this so that the `basePath` *could* contain a trailing slash (if one is present in the user config) and only trim it to a single slash (if present), otherwise continue without a trailing slash?

packages/config/index.js
```diff
@@ -40,14 +74,13 @@ function resolvePaths(config) {
...
   var assetPath = config.assetPath.replace(/(^\/*|\/*$)/g, '');
-
-  return Object.assign(config, {
+  return Object.assign({}, config, {
     locations: config.locations
       .map(function(location) {
         return basePath + location.replace(/\/*$/, '').replace(/^\/*/, '/');
```
ZauberNerd: Same as above - we should allow optional trailing slashes.

packages/config/index.js
```diff
@@ -4,34 +4,68 @@ var path = require('path');
...
-    require.resolve(path.join(root, 'package.json'));
-    npmConfig = require(path.join(root, 'package.json')).config.hops;
-  } catch (_) {}
+var explorer = require('cosmiconfig')('hops', {
+  rcExtensions: true,
+  stopDir: root,
+  sync: true,
+});
+
+function getRawConfig() {
```
ZauberNerd: I think this name is a bit misleading, why not split it in `getDefaultConfig` and `getUserConfig` or similar?

packages/config/index.js
ZauberNerd: We could also use `resolve-cwd` or one of its related projects: https://www.npmjs.com/package/resolve-cwd#related

packages/config/index.js
```diff
@@ -4,34 +4,68 @@ var path = require('path');
...
+      cacheDir: 'node_modules/.cache/hops',
+    },
+    result ? result.config : {}
+  );
 }
 
-function extendConfig(config) {
-  if (npmConfig.extends) {
+function applyExtension(config) {
+  var result = Object.assign({}, config);
```
matthias-reis: Gut feeling: we should generally opt in for a deep merge.

Let's say, we provide sane defaults for deployment configs in production (under `env`), how could you overwrite single entries here?

packages/config/README.md
dmbch: I think I agree with @matthias-reis's [gut](https://github.com/xing/hops/pull/336#discussion_r161482470) - I think I'd like to go with `deep-merge`.

packages/config/README.md
dmbch: The extension mechanism is already documented - I think what is there should do for now.

packages/config/index.js
dmbch: I have now renamed  it to `applyExtensionConfig()` - does that work for you?

packages/config/index.js
dmbch: While we certainly could, I do not really see any benefit, here - but that would imply touching a whole lot of different places (react-router config springs to mind).

packages/config/index.js
dmbch: For the asset path, I fail to see a good reason for a change. I feel we only normalise a URL path that is only ever used internally to build asset URLs. Why should we change that mechanism?

packages/config/index.js
dmbch: True. I somehow also feel we might have to go for `deep-assign`.

packages/config/index.js
dmbch: I concur - I think I will switch from `Object.assign()` to [deep-assign](https://www.npmjs.com/package/deep-assign)

packages/config/index.js
dmbch: I would not want to get rid of that mechanism altogether as that would force basically everyone to add a JS config file to their projects - at least internally.

On the other hand, it appears to me that inheriting from multiple config modules is a rather advanced use-case. It such cases, I would consider it acceptable to have a JS config. Do you agree?

matthias-reis: You're right. Let's keep it simple...

ZauberNerd:
Please do a search through all files in the project for `"config"`, there are still some more references to the "npm style" config:
* [packages/build-config/README.md](https://github.com/xing/hops/blob/master/packages/build-config/README.md)
* [packages/graphql/README.md](https://github.com/xing/hops/blob/master/packages/grahpql/README.md)
* [packages/lambda/README.md](https://github.com/xing/hops/blob/master/packages/lambda/README.md)
* [packages/template-graphql/README.md](https://github.com/xing/hops/blob/master/packages/template-graphql/README.md)
* [packages/template-react/README.md](https://github.com/xing/hops/blob/master/packages/template-react/README.md)

packages/config/README.md
```diff
@@ -15,7 +15,7 @@ npm install --save hops-config
 
 # Usage
 
-Please find a list of the default options below. They can be set in your project's package.json and can be overridden by using [npm config](https://docs.npmjs.com/cli/config) in the command line. You can extend hops-config within your project allowing you to use it to configure your isomorphic applications. The configuration is immutable at runtime - if you import it in your own application, you can only read its values, but you cannot alter them.
+Please find a list of the default options below. They can be set in your project's package.json. You can extend hops-config within your project allowing you to use it to configure your isomorphic applications.
```
ZauberNerd: As this is the first mention of the usage, I'd probably write something like: "They can be set in your project's package.json (or dedicated files - see below)."

packages/config/README.md
```diff
@@ -43,57 +43,68 @@ Please find a list of the default options below. They can be set in your project
...
+2. `.hopsrc` (or `.hopsrc.(js|json|ya?ml)`)
+3. `hops.config.js` JS file exporting the object
 
-```
-npm config set my-application:hops:port 1337
+You can use whatever suits you best, but please only use a single location in your project's root folder. Hops does not attempt to merge configurations stored in different locations: it uses the first config it finds.
+
+## Environment Specific Configuration
+
+Take the above package.json - and let's say you want to want to override the `port` value. To override the port value for your local environment without changing your application's overall config, add an environment specific config...
```
ZauberNerd: We should mention, that this uses the environment variable `HOPS_ENV` and falls back to `NODE_ENV`.

packages/config/index.js
```diff
@@ -2,36 +2,77 @@
...
+    envVars: { HOPS_MODE: 'dynamic' },
+    moduleDirs: [],
+    appDir: '.',
+    buildDir: 'build',
+    cacheDir: 'node_modules/.cache/hops',
+  };
+}
+
+function applyUserConfig(config) {
+  var result = explorer.load(process.cwd());
```
ZauberNerd: What's the reason for using `process.cwd()` here instead of `root` (which points to the hops application root)?

packages/config/README.md
ZauberNerd: I'm having trouble finding where the extension mechanism is documented - can you point me to it?

packages/config/index.js
ZauberNerd: Ok, let's keep it out of this PR and I'll think about it a bit more and create an issue for this.

packages/config/index.js
dmbch: I used it because it most closely resembles the present implementation. I can imagine some (rather weird) monorepo setups in which this might be helpful. I many cases both approaches produce identical values, I guess.

Issue #342, opened 2018-01-19
Import bulma css


ShintaroNippon:
## How to import bulma

i installed bulma then i import in app js
but im getting an error...

node_modules/bulma/bulma.sass:1
(function (exports, require, module, __filename, __dirname) { @charset "utf-8"

HerrBertling:
Looking at your error message, you apparently try to import a `.sass` file – I guess this happens when you `import bulma from 'bulma'`? The `main` entry in the `bulma` `package.json` links to that file: https://github.com/jgthms/bulma/blob/master/package.json#L7

Unfortunately, Hops currently only supports pure CSS files and PostCSS with [cssnext](https://github.com/MoOx/postcss-cssnext) at the moment.

I don't think there are any plans to extend this to include SASS files.

You could however import the `bulma.css` file instead which should work without problems.

PR #343, opened 2018-01-19
Fix CSS Imports

dmbch:
This pull request closes issue #342

## Current state

Previously, most of `node_modules` was excluded from bundling in Node.js builds. Unfortunately, this also prevented CSS files (and other file types) from being imported.

## Changes introduced here

This commit adds an RegExp check to make sure we bundle all file types except JSON and JS.

## Checklist

### DIFF
--- a/packages/build-config/configs/node.js
+++ b/packages/build-config/configs/node.js
@@ -28,7 +28,7 @@ function shouldIncludeExternalModuleInBundle(module) {
     module.indexOf('babel-polyfill') === 0 ||
     // bundle everything the native require implementation can not handle
     // see: https://nodejs.org/api/modules.html#modules_all_together
-    /^.*\.(?!(?:js|json|mjs|node)$)/.test(module) ||
+    !/\.(?:js|json|mjs|node)$/.test(require.resolve(module)) ||
     checkEsnext(module)
   );
 }
--- a/packages/build-config/configs/node.js
+++ b/packages/build-config/configs/node.js
@@ -26,7 +26,9 @@ function shouldIncludeExternalModuleInBundle(module) {
     module.indexOf('hops') === 0 ||
     module.indexOf('core-js') === 0 ||
     module.indexOf('babel-polyfill') === 0 ||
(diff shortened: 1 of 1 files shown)


DJCordhose:
I can not comment on whether only excluding js and json files is the right approach (maybe not), but probably the regexp is a bit too complicated and thus hard to maintain. Why  not js|json?

DJCordhose:
Also, what about other files types we do not want included? Like e.g. d.ts? Maybe rather use a white list?

packages/build-config/configs/node.js
```diff
@@ -23,9 +23,10 @@ function findNodeModules(start) {
 
 function shouldIncludeExternalModuleInBundle(module) {
   return (
+    module.indexOf('hops') === 0 ||
     module.indexOf('core-js') === 0 ||
     module.indexOf('babel-polyfill') === 0 ||
-    module.indexOf('hops') === 0 ||
+    /^.*\.(?!js(?:on)?$)/.test(module) ||
```
ZauberNerd: We should also add `.mjs` and maybe even `.node`.
Also there should be a comment, explaining, that this is based on the Node.js resolve algorithm: https://nodejs.org/api/modules.html#modules_all_together

packages/build-config/configs/node.js
```diff
@@ -23,9 +23,12 @@ function findNodeModules(start) {
 
 function shouldIncludeExternalModuleInBundle(module) {
   return (
+    module.indexOf('hops') === 0 ||
     module.indexOf('core-js') === 0 ||
     module.indexOf('babel-polyfill') === 0 ||
-    module.indexOf('hops') === 0 ||
+    // bundle everything the native require implementation can not handle
+    // see: https://nodejs.org/api/modules.html#modules_all_together
+    /^.*\.(?!(?:js|json|mjs|node)$)/.test(module) ||
```
ZauberNerd: This breaks for `.`, `./foo`, `../../foo` and `lodash.find` etc.
We could try to `require.resolve` the `module` before.

Issue #324, opened 2017-12-22
hops-graphql: getDataFromTree may not work with all kinds of apps


Xiphe:
## This is my intent (choose one)

- [x] I want to report a bug
- [x] I want to request a feature or change
- [ ] I want to provide or change a feature

#### Observation

In the past I used apollo client without `hops-graphql` and just with the basic apollo features and everything worked great.
Now I try to migrate to `hops-graphql` and suddenly my ssr does not work anymore because values in react context are missing.

#### Thesis

It seems as if the rendering routine in [`getDataFromTree`](https://www.apollographql.com/docs/react/recipes/server-side-rendering.html#getDataFromTree) does not work well with `React.cloneElement` and/or Reacts context API.

#### Debugging

I disabled [ssr mode](https://www.apollographql.com/docs/react/recipes/server-side-rendering.html#skip-for-ssr) and tried to find the root cause for this but failed and will now give up after two hours.

I also failed to isolate the problem in a reproducible state.

## Proposed solution

In our use-case we do not need graphql queries to be run on the server since the data is highly user-specific and we use static pre-build-ssr and not dynamic ssr.

Therefore I propose to add an option to disable the ssr pre-fetching.

dmbch:
Thanks for your report - and your patience! It would be great if you could provide a simple test case for us to investigate. Lacking that, we probably will not be able to be of a lot of assistance with debugging, but we will investigate your proposal (i.e. global `ssr` flag).

That being said, you could probably extend `GraphQLContext` and override the default [`getTemplateData`](https://github.com/xing/hops/blob/98b1c791c1528091e6667e0b0dd0717b49615e32/packages/graphql/node.js#L23) method. Would that help in your case?

MikeBild:
Is a small modification like:

https://github.com/xing/hops/blob/98b1c791c1528091e6667e0b0dd0717b49615e32/packages/graphql/node.js#L17

```javascript
ssrMode: common.ssrMode || true
```

https://github.com/xing/hops/blob/98b1c791c1528091e6667e0b0dd0717b49615e32/packages/graphql/lib/common.js#L23

```javascript
  enhanceClientOptions: function(options) {
    return Object.assign({}, options, {
      link: options.link || this.createLink(),
      cache: options.cache || this.createCache(),
      ssrMode: options.ssrMode
    });
  },
```

on option?

Xiphe:
[...]

PR #344, opened 2018-01-20
Disable GraphQL Data Fetching in Static Mode

dmbch:
This pull request closes issue #324 

## Current state

Currently, GraphQL data is always being prefetched, even in static rendering mode.

## Changes introduced here

In static mode, during server side rendering, GraphQL data is no longer being prefetched. That appears to be the sane default. The old behaviour can be restored by extending `GraphQLContext` and overriding its 'prefetchData' method.

## Checklist


robin-drexler:
❤️

PR #349, opened 2018-01-23
build: add prerelease scripts

dmbch:


ZauberNerd:
We could also add `"exact": true` to the lerna.json under the `commands.publish` section, as it is used for every invocation of the publish command.
Also: `--force-publish=*` should only be needed for `major` and `premajor`, right?

dmbch:
@ZauberNerd PTAnotherL

PR #350, opened 2018-01-24
Add Header component to allow setting http headers declaratively

robin-drexler:
## Current state
No way to set http headers from inside the application.
This is useful for example to set (surrogate) cache headers or to set cookies. 

## Changes introduced here
Header component is added to allow setting http headers

**Needs tests.**

## Checklist

* [x] Documentation has been added


robin-drexler:
I wonder if the renderer should (also) resolve with the entire response object instead of only the [rendered body].(https://github.com/xing/hops/blob/e7536e9ae8fa080db7ab46b3974b4cb468c06b6a/packages/renderer/index.js#L40)

This would allow clients using `hopsConfig.generate` to also process the headers and tell a backend to respect cache headers. (e.g. by generating a [firebase config](https://firebase.google.com/docs/hosting/full-config))

PR #353, opened 2018-01-25
New Dev Server

dmbch:
## Current state

We are using webpack-dev-server in development mode, but do not use any of its advanced features. This also implies that we are not using the exact same version of Express in development and in production.

## Changes introduced here

We are now using our own Express server in development - the same as in production, but in slightly different setups (security, caching etc. differ, of course).

Another benefit of this change is that Hops now also supports SSL in production mode.

## Checklist


dmbch:
Will be dealt with in #354

PR #355, opened 2018-01-30
feat(react): render variant hiding contexts from users

DJCordhose:
**Proposal for a simplified alternative to the existing react based rendering.**

Objective: hide context (a truly advanced feature) from the react developer

Instead of passing a context to render, you can now pass a configuration object that also describes which additional rendering features you need, like:

 ```javascript
import { reduxExtension } from 'hops-redux';

export default render(<App />, {
  extensions: [reduxExtension({ reducers })],
});
```

Each react based package now exports itself as an extension.

**This is a non-breaking change, if you want to you can still pass a context.**

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -29,6 +29,25 @@ const App = () => <h1>Hello World!</h1>;
 export default render(<App />, createContext());
 ```
 
+## `render(reactElement, config)`
+
+Simple wrapper for [`render(reactElement, context)`](#renderreactelement-context) shielding the details of context from you. Each package that functions as an optional extension to the core react package offers a so called `extension` which you can pass as an array. Unlike [`React#createContext(options)`](#createcontextoptions) each extension only takes a configuration that is specific to it, not a general one.
+
+Here you have an example of a React application using the Redux extension. The Redux extension takes only the specific set of options (not the general React ones) mentioned in [`Redux#createContext(options)`](../redux/README.md#createcontextoptions). The general options are passed directly in the configuration.
+
+```js
+import React from 'react';
+import { render } from 'hops-react';
+import { reduxExtension } from 'hops-redux';
+
+const App = () => <h1>Hello World!</h1>;
+
+export default render(<App />, {
+  mountpoint: '#main',
+  extensions: [reduxExtension({ reducers })],
+});
+```
+
 ## `createContext(options)`
 
 `createContext()` generates a rendering context containing most of the actual implementation used by `render`. It takes a couple of options:
--- a/packages/react/dom.js
+++ b/packages/react/dom.js
@@ -7,6 +7,8 @@ var mixinable = require('mixinable');
 
 var hopsConfig = require('hops-config');
 
+var createCombinedContext = require('./lib/common').createCombinedContext;
+
 exports.combineContexts = mixinable({
   bootstrap: mixinable.async.parallel,
   enhanceElement: mixinable.async.compose,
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: packages/react/dom.js [hunk 1]
```diff
@@ -67,4 +69,18 @@ exports.render = function(reactElement, context) {
   };
 };
 
+exports.render = function(reactElement, contextOrConfig) {
+  var context;
+  if (typeof contextOrConfig === 'function') {
```
dmbch: I am afraid we are going to need a different check here - `contextOrConfig` could be a context instance here, but I would not really expect a function/constructor.

packages/react/node.js
```diff
@@ -109,4 +111,18 @@ exports.render = function(reactElement, _context) {
   };
 };
 
+exports.render = function(reactElement, contextOrConfig) {
+  var context;
+  if (typeof contextOrConfig === 'function') {
```
dmbch: Same here - `contextOrConfig` should not really be a function...

packages/react/lib/common.js
```diff
@@ -0,0 +1,21 @@
...
+) {
+  var mergedConfig = Object.assign({}, config);
+  delete mergedConfig.extensions;
+  var mergedContexts = [ReactContext];
+  if (config.extensions) {
+    config.extensions.forEach(function(extension) {
+      var context = extension.context;
+      mergedContexts.push(context);
+      var extensionConfig = extension.config;
+      mergedConfig = Object.assign(mergedConfig, extensionConfig);
```
dmbch: I think I would rather `bind` the extension specific options to their respective constructor, something like..:

```
var config = Object.assign({}, mergedConfig, extensionConfig);
mergedContexts.push(
  Function.prototype.bind.call(context, context, config)
);
```

dmbch: ...but that is clearly suboptimal, as well (cloning...)

packages/react/dom.js
dmbch: I think we ought to use the brand new [`isMixinable()`](https://www.npmjs.com/package/mixinable#defineismixinableobject) function here - and we could additionally check whether we have a proper `Context`.

packages/react/lib/common.js
dmbch: Meanwhile, I have come to think that your initial implementation is the best we can do without changing the way `mixinable` works.

packages/react/README.md
```diff
@@ -29,6 +29,25 @@ const App = () => <h1>Hello World!</h1>;
 export default render(<App />, createContext());
 ```
 
+## `render(reactElement, config)`
```
dmbch: If we consider this the 'consumer' API, i.e. the most approachable way to use `render()`, why do we not mention it first? I really like this API and I think we ought to expose it most prominently.

packages/react/dom.js
```diff
@@ -40,7 +42,7 @@ exports.contextDefinition = exports.ReactContext;
 
 exports.createContext = exports.combineContexts(exports.ReactContext);
 
-exports.render = function(reactElement, context) {
+exports.renderWithContext = function(reactElement, context) {
```
dmbch: Why do we export `renderWithContext`? Can we not just use `render`?

packages/react/README.md
DJCordhose: I had the same thoughts, but I had to somehow refer to the original render function for motivation, so it had to come first, I thought. Maybe not a valid point, not sure...

packages/react/dom.js
DJCordhose: Yes, was broken, works fine with `isMixinable` :)

DJCordhose:
As this is a non breaking change and commits are now squashed, I think it might be time for a merge if you want it. Details can be worked on once merged.

PR #373, opened 2018-02-05
fix(config): deep-merge config without array concatenation

ZauberNerd:


packages/config/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "dependencies": {
     "cosmiconfig": "^3.1.0",
+    "lodash.mergewith": "^4.6.1",
     "deep-assign": "^2.0.0",
```
dmbch: I think we can get rid of that one now, can we not..?

packages/config/index.js
```diff
@@ -3,11 +3,21 @@
...
 var path = require('path');
 
-var assign = require('deep-assign');
+var mergeWith = require('lodash.mergewith');
 var root = require('pkg-dir').sync();
 
 var cosmiconfig = require('cosmiconfig');
 
+function assign(target) {
+  var args = [];
```
dmbch: why not just `var args = Array.prototype.slice.call(arguments, 1);`?

PR #374, opened 2018-02-05
Placeholders Everywhere

dmbch:
## Current state

Currently, placeholders are not resolved in plain objects used within our configs.

## Changes introduced here

Now they are - just in plain objects, though!

## Checklist

### DIFF
--- a/packages/config/index.js
+++ b/packages/config/index.js
@@ -3,10 +3,11 @@
 var fs = require('fs');
 var path = require('path');
 
-var mergeWith = require('lodash.mergewith');
-var root = require('pkg-dir').sync();
-
 var cosmiconfig = require('cosmiconfig');
+var mergeWith = require('lodash.mergewith');
+var isPlainObject = require('is-plain-object');
+
+var root = require('pkg-dir').sync();
 
 function assign() {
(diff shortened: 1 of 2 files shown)


packages/config/index.js
```diff
@@ -81,24 +82,29 @@ function applyEnvironmentConfig(config) {
...
+        return result;
+      }, {});
+    }
     if (typeof item === 'string') {
-      return item.replace(regExp, function(_, match) {
-        var result = config[match].toString();
+      return item.replace(regExp, function(_, configKey) {
+        var result = config[configKey];
+        if (typeof result !== 'string') {
+          result = String(result).toString();
```
ZauberNerd: What is the reason for `String(x).toString()` instead of just `String(x)`?

dmbch: I just wanted to make it obvious that we are dealing with a primitive string here - I always found it a bit confusing that..:

```js
'foo' === 'foo'
true
String('foo') === String('foo')
// true
String('foo') === new String('foo')
// false
```

But maybe that's just me..?

dmbch: (of course I have come to terms with what is going on in there, but anyway...)

PR #378, opened 2018-02-08
test(spec): re-enable non-config tests

dmbch:


robin-drexler:
Ooops, nice catch and sorry. This could use a linter rule.

PR #390, opened 2018-02-12
chore: exclude **/package.json files from prettier formatting

ZauberNerd:
eslint does not lint *.{json,css,md} files and therefore does not catch
differences in prettier formatting in these files.

~This commit adds a separate command that executes "prettier -l" to list
the differences in the above mentioned files and fails the "lint"
script if there are any differences reported~

We are now excluding `**/package.json` files from prettier formatting.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,5 @@
+node_modules/
+packages/*/**/build
+package-lock.json
+fragmentTypes.json
+**/package.json
--- a/package.json
+++ b/package.json
@@ -1,7 +1,11 @@
 {
   "name": "hops",
   "description": "universal build tools",
-  "keywords": ["webpack", "babel", "postcss"],
+  "keywords": [
+    "webpack",
+    "babel",
+    "postcss"
+  ],
   "private": true,
   "license": "MIT",
   "homepage": "https://github.com/xing/hops",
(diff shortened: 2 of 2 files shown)


ZauberNerd:
After thinking about this a bit more, I think it won't work in our current setup, as `yarn`, `npm` and perhaps `renovate` too will rewrite the `package.json` files and therefore always cause linting issues.

Solution: Exclude `**/package.json` files from prettier formatting
Problem: We currently use the `.gitignore` file as `prettier --ignore-path .gitignore` but we can't gitignore `package.json` files.
We could either wait for: https://github.com/prettier/prettier/pull/3737 or we duplicate the entries from the `.gitignore` file into a separate `.prettierignore` file in which we also exclude the `package.json` files.

robin-drexler:
Couldn't we use `prettier -l` for all files and get rid of the eslint plugin/rules?

dmbch:
@robin-drexler I think we ought to combine the two as eslint catches issues prettier does not cover. But we could get rid of the eslint/prettier integration.

Although..: That would potentially break tooling/editor integration.

ZauberNerd:
@robin-drexler @dmbch is correct - we should keep the prettier/eslint integration, the `eslint-config-prettier` for **disabling all conflicting rules** and the `eslint-plugin-prettier` to **automatically fix issue** when using `eslint --fix` as that is something that, for example the vscode quick-fix section uses.
I think that for now this solution is the best we can come up with - until the above mentioned prettier PR gets merged or we decide to duplicate our `.gitignore` to a `.prettierignore` file.

ZauberNerd:
So, it turns out that lerna bypasses all git hooks when committing: https://github.com/lerna/lerna/blob/8fb38e4102daa5ff2526e0b2604a6ee8e6dd26f8/src/GitUtilities.js#L53 and it does not have an option to configure this behavior.

That means that the "lint-fmt" script in this PR will never get executed when publishing - therefore we will always have unformatted `package.json` files on master after a publish.

- We could either try to contribute a flag to lerna to enable git hooks when publishing or
- we could ignore all `package.json` files from lerna formatting.

I'd prefer the second option.
That would mean we have to create a separate `.prettierignore` file and duplicate the contents of the `.gitignore` file in there - but since the `.gitignore` file rarely changes, I think it won't be a big burden.

Issue #319, opened 2017-12-19
aws config leaks into client bundle 


robin-drexler:
## This is my intent (choose one)

- [x] I want to voice a concern
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem
The entire aws config can be found in the main.js because it is part of hops-config which gets inlined into the bundle.
This includes possibly confidential data such as certificateArn or information about the projects folder structure. 

## Proposed solution
If it is agreed upon that this rather suboptimal, we could allow to blacklist keys in hops-config that will not be part of the client bundle. `aws` could be one of the defaults.

dmbch:
@ZauberNerd would you mind chiming in..? ;)

ZauberNerd:
With the new config mechanism: https://github.com/xing/hops/commit/af402cea12c0e28270e260f145e237f62862bf26 we could just rename the aws config into `_aws` and be done with it, no?
In any case, I'm not sure that the information that can currently be accessed through `hops-config` would be considered confidential: https://github.com/xing/hops/tree/master/packages/lambda#available-options-through-hops-npm-config

dmbch:
Yes, you are completely right: prefixing with `_` will indeed exclude the AWS config from the browser build - and the the data currently exposed does not appear to be sensible.

I think using the `_aws`-approach is the Right Thing (TM), anyway.

PR #392, opened 2018-02-12
feat(lambda): hide aws configuration from browser by prefixing with _

ZauberNerd:
This commit changes the lambda configuration to use the key `_aws`
instead of `aws` which means the aws config will not be included in
the client-side JS bundle.

If `hopsConfig._aws` is not found it falls back to `hopsConfig.aws` to
ensure that this is a non-breaking backwards-compatible change.

Fixes: #319

### DIFF
--- a/packages/lambda/README.md
+++ b/packages/lambda/README.md
@@ -13,16 +13,12 @@ This is the minimal configuration required to get started with hops-lambda:
 ```diff
 --- a/package.json
 +++ b/package.json
-@@ -6,5 +6,8 @@
-   "config": {
-     "hops": {
-+      "assetPath": "prod",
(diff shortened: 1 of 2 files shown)


ZauberNerd:
@robin-drexler the `config` tests don't have anything to do with `hops-lambda` and `aws` in this test case is just an arbitrarily chosen name for a nested object.
Updating these tests will not verify anything, because this PR addresses this file (https://github.com/xing/hops/blob/f24133f2e0ce0911a4cfe8aab386c88456fb413b/packages/lambda/lib/aws-config.js) in the `hops-lambda` package.

robin-drexler:
True, don't mind me then :)

PR #398, opened 2018-02-13
feat(express): statically serve app shells for dynamic path segments

ZauberNerd:
This allows us to define dynamic path segments in the `locations`
section of the `hops-config` the same way as in our route definitions.

**Example:**
for the following routes (defined in your application):
```javascript
<Switch>
  <Route exact path="/categories" component={Categories} />
  <Route exact path="/exercises/:category" component={Exercises} />
</Switch>
```
you can configure these locations to be pre-rendered as app shells:
```json
  "hops": {
    "locations": ["/categories", "/exercises/:category"]
  }
```

And when executing `hops build --static` the above defined route
component (`Exercises`) will receive:
`props.match.category = ':category'`
to enable you to decide whether to render an app shell or actual
content.

When using `hops-express` to serve static routes only
(`hops serve --static`) it will automatically find the matching
app shells for the requested route.

### DIFF
--- a/packages/build/lib/generate.js
+++ b/packages/build/lib/generate.js
@@ -14,7 +14,9 @@ var createRenderer = require('./renderer');
 function getFileName(location) {
   return path.join.apply(
     path,
-    [hopsConfig.buildDir].concat(index(location).split(path.sep))
+    [hopsConfig.buildDir].concat(
+      index(location.replace(/:/g, '')).split(path.sep)
+    )
   );
 }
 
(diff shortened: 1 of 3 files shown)


dmbch:
Generally, I would prefer using [`minimatch`](https://github.com/isaacs/minimatch): Seeing `react-router`-style patterns in `locations` would make me wonder what happens to those named matches...

ZauberNerd:
Closing this for now - we can revisit it at a later time if it is still required.

PR #399, opened 2018-02-13
Rewrite code base to ES6

ZauberNerd:
The commits in this branch are atomic and it is therefore recommended to look at the change-set by browsing through the commits instead of looking at the entire "Files changed" section.


ZauberNerd:
Tests are broken because of: https://github.com/xing/hops/commit/5786b23d9b250afce9a1510e374996216dcee962
We could either revert that commit or wait for: https://github.com/mysticatea/eslint-plugin-node/pull/109

PR #400, opened 2018-02-14
Alternative approach to Service Workers

ZauberNerd:
This is an alternative approach to #375

### DIFF
--- a/packages/build-config/plugins/webpack-service-worker.js
+++ b/packages/build-config/plugins/webpack-service-worker.js
@@ -1,0 +1,56 @@
+'use strict';
+
+var SingleEntryPlugin = require('webpack/lib/SingleEntryPlugin');
+var RawSource = require('webpack-sources').RawSource;
+
+var hopsConfig = require('hops-config');
+
+var PLUGIN_NAME = 'hops-service-worker-plugin';
+
+module.exports = function ServiceWorkerPlugin() {
+  var assetPath = hopsConfig.workerPath.replace(/^\/+/, '');
+
+  this.apply = function(compiler) {
+    if (!hopsConfig.workerFile) {
+      return;
+    }
+
+    function onMake(compilation, callback) {
+      compilation
+        .createChildCompiler(PLUGIN_NAME, { filename: assetPath }, [
+          new SingleEntryPlugin(
+            compiler.context,
+            require.resolve('../shims/worker-shim'),
+            'worker'
+          ),
+        ])
+        .runAsChild(callback);
+    }
+
+    function onEmit(compilation, callback) {
+      compilation.assets[assetPath] = new RawSource(
+        'HOPS_ASSETS = ' +
+          JSON.stringify(
+            Object.keys(compilation.assets).filter(function(item) {
+              return !item.match(/\.map|stats\.json$/) && item !== assetPath;
+            })
+          ) +
+          ';\n' +
+          compilation.assets[assetPath].source()
+      );
+      callback();
+    }
+
+    if (typeof compiler.hooks !== 'undefined') {
+      compiler.hooks.make.tapAsync(PLUGIN_NAME, onMake);
+    } else {
+      compiler.plugin('make', onMake);
+    }
+
+    if (typeof compiler.hooks !== 'undefined') {
+      compiler.hooks.emit.tap(PLUGIN_NAME, onEmit);
+    } else {
+      compiler.plugin('emit', onEmit);
+    }
+  };
+};
--- a/packages/build-config/sections/module-rules/webmanifest.js
+++ b/packages/build-config/sections/module-rules/webmanifest.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var path = require('path');
+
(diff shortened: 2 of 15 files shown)


packages/build-config/sections/module-rules/webmanifest.js
```diff
@@ -0,0 +1,20 @@
...
+
+var hopsConfig = require('hops-config');
+
+exports.default = {
+  test: /\.webmanifest$/,
+  use: [
+    {
+      loader: require.resolve('file-loader'),
+      options: {
+        name: path.join(hopsConfig.assetPath, '[name]-[hash:16].[ext]'),
```
ZauberNerd: Should we add `emitFile: false` here too?

packages/build-config/shims/worker-shim.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+require('babel-polyfill');
+
+(function execute() {
+  var entryPoint = require('hops-worker-entry-point');
+  if (entryPoint.__esModule) {
```
ZauberNerd: Should probably also be `typeof entryPoint !== 'function' && typeof entryPoint.default === 'function'`

packages/build-config/shims/worker-shim.js
```diff
@@ -0,0 +1,16 @@
...
+    entryPoint = entryPoint.default;
+  }
+  entryPoint(HOPS_ASSETS); // eslint-disable-line no-undef
+  if (module.hot) {
```
ZauberNerd: Not sure how Service Workers behave with hot module reloading.. needs more investigation.

packages/pwa/dom.js
```diff
@@ -0,0 +1,26 @@
...
+module.exports = function installServiceWorker() {
+  return new Promise(function(resolve, reject) {
+    if (!('serviceWorker' in navigator)) {
+      return reject(
+        new Error('ServiceWorkers are not supported in this browser')
+      );
+    }
+
+    if (
+      window.location.protocol === 'https' ||
```
ZauberNerd: Should we have Service Workers in development and production builds? Or do we need to treat these differently?

packages/pwa/dom.js
```diff
@@ -0,0 +1,26 @@
...
+      window.location.hostname === 'localhost'
+    ) {
+      window.addEventListener('load', function() {
+        resolve(
+          navigator.serviceWorker.register(hopsConfig.workerPath, {
+            scope: hopsConfig.basePath,
```
ZauberNerd: Should we always use the `basePath` as scope or are there cases, where we would like to specify a scope that is different from the `basePath`?

packages/pwa/node.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = function installServiceWorke() {
+  return new Promise(function() {});
```
ZauberNerd: Should add a comment to explain that this promise intentionally never resolves. (Can't register a Service Worker in Node.js environment).

packages/build-config/plugins/webpack-service-worker.js
```diff
@@ -0,0 +1,56 @@
...
+            compiler.context,
+            require.resolve('../shims/worker-shim'),
+            'worker'
+          ),
+        ])
+        .runAsChild(callback);
+    }
+
+    function onEmit(compilation, callback) {
+      compilation.assets[assetPath] = new RawSource(
```
ZauberNerd: Maybe this should also contain `locations`? Or we fetch `locations` inside Service Worker directly by importing `hopsConfig.locations`.

packages/pwa/dom.js
dmbch: As `basePath` is the actual application's root path, I think we do not have to cater for this.

packages/build-config/sections/module-rules/webmanifest.js
dmbch: Yes, please. And in node, we could probably also get rid of the `web-app-manifest-loader` itself.

packages/build-config/plugins/webpack-service-worker.js
robin-drexler: +1 for hopsConfig. It's how you'd access locations anywhere else too, so I guess it make sense to have it consistent.

packages/build-config/shims/worker-shim.js
robin-drexler: Shall we disable it for now?

packages/config/index.js
```diff
@@ -29,6 +29,8 @@ function applyDefaultConfig(config) {
       locations: [],
       basePath: '',
       assetPath: '',
+      workerPath: '<assetPath>/sw.js',
```
dmbch: please replace `<assetPath>` with `<basePath>` to make sure the SW can control the actual app

packages/pwa/dom.js
dmbch: We _do_ support SSL in both prod and dev environments, so I would prefer keeping things simple, here.

PR #404, opened 2018-02-15
General refactorings (taken from webpack@4 branch)

ZauberNerd:


HerrBertling:
General Refactorings

packages/build/commands/build.js
```diff
@@ -42,10 +42,5 @@ module.exports = function defineBuildCommand(args) {
 };
 
 if (require.main === module) {
-  try {
```
ZauberNerd: I'd prefer to leave these checks in for now and log a deprecation warning with instructions on how to fix it because the change of `hops-cli`/`hops-local-cli` to `hops` was also a non-breaking change and might not have reached all teams using hops.
We can then remove these checks and warnings in an upcoming major release.

PR #410, opened 2018-02-19
fix(express): don't cache sw.js

ZauberNerd:


packages/express/lib/app.js
```diff
@@ -10,6 +10,8 @@ var helmet = require('helmet');
 var hopsConfig = require('hops-config');
 var utils = require('./utils');
 
+var swRe = new RegExp(hopsConfig.basePath + '/sw.js$');
```
robin-drexler: This should probably be `hopsConfig.workerPath`, should it not?

PR #412, opened 2018-02-19
feat(express): enable gzip compression

ZauberNerd:
This PR adds gzip compression to the express server.

~WIP because it needs investigation how Lambda / API Gateway behave when
they deal with gzip compressed responses.~
API Gateway expects Lambda to return compressed responses in order to serve compressed responses to the client.

### DIFF
--- a/packages/express/lib/app.js
+++ b/packages/express/lib/app.js
@@ -6,6 +6,7 @@ var path = require('path');
 var express = require('express');
 var mime = require('mime');
 var helmet = require('helmet');
+var compression = require('compression');
 
 var hopsConfig = require('hops-config');
 var utils = require('./utils');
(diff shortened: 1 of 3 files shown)


robin-drexler:
💯🔥

ZauberNerd:
Lambda / API Gateway does not do compression on their own, but will handle gzip compressed responses properly when the Lambda function returns a gzipped response.
Therefore I think we should implement this. If there really is concern about performance in other scenarios we could think about enabling compression only when NODE_ENV is set to "production" or even rely on other environment variables that are only set on AWS Lambda.

ZauberNerd:
@dmbch we could for example check if the `LAMBDA_TASK_ROOT` is set, which should only happen on AWS Lambda, so that Hops running behind a reverse proxy will still serve uncompressed content and leave the compression to the proxy: https://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html#lambda-environment-variables

dmbch:
@zaubernerd interesting idea. Although we could also merge this feature as-is and iterate..?

N.B.: yeah, we changed my mind. And we must make this stuff less monolithic :grin:

PR #432, opened 2018-03-12
allow function as Header prop

robin-drexler:
## Current state
Header component can only be used to override headers. 

## Changes introduced here
`value` can be a function which gets passed headers that were already set, which allows to derive headers from those.
E.g. one could add to existing cookies.

## Checklist

* [ ] All commit messages adhere to the [appropriate format](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit) in order to trigger a correct release
* [ ] All code is written in plain ECMAScript v5 and is formatted using [prettier](https://prettier.io)


robin-drexler:
This just came to my mind. Not sure if it's really necessary, though. :D

PR #433, opened 2018-03-15
Generate fragment types from local schema file

ZauberNerd:
When providing Hops with a GraphQL schema file the command
`hops graphql introspect` will try to generate the fragment types
(required for querying fragments on unions or intersections) from the
provided file. If this file does not exist, it will try to generate the
fragment types by querying the GraphQL server specified via "graphqlUri"


packages/graphql/__tests__/graphql.spec.js
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  it('should create and configure the apollo client', function() {
+    var context = new hopsGraphql.GraphQLContext();
+    expect(context.client).toBeDefined();
+  });
+
+  it('should allow to override apollo link through options', function() {
+    var mockLink = {};
+    // eslint-disable-next-line
```
robin-drexler: Could you add which rule should be suppressed?

packages/graphql/__tests__/node.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    hopsGraphql = require('../node');
+  });
+
+  it('should set ssrMode', function() {
+    var context = new hopsGraphql.GraphQLContext();
+
+    expect(context.client.ssrMode).toBe(true);
+  });
+
+  it.only('should not prefetch in static mode', function() {
```
robin-drexler: `.only` needs to be removed.
For mocha we had an eslint rule in place to catch such errors. Maybe we should do the same for jest. :)

packages/graphql/lib/fragments.js
```diff
@@ -1,49 +1,66 @@
...
-module.exports = function fetchFragments() {
-  return fetch(hopsConfig.graphqlUri, {
+function writeFragmentTypesFile(result) {
+  result.data.__schema.types = result.data.__schema.types.filter(function(t) {
+    return t.possibleTypes !== null;
+  });
+  return pify(fs.writeFile)(fragmentsFile, JSON.stringify(result.data));
+}
+
+function executeRemoteQuery(graphqlUri, query) {
```
robin-drexler: Some users might need to add additional headers/params to authenticate with their graphQL endpoint.

Do you think it'd be helpful to allow that in any way?

PR #438, opened 2018-03-16
[WIP]fix: do not use loader name options to configure assetPath anymore

robin-drexler:
For a specific image I'd like to prevent url-loader from returning the base64 version of an image regardless of its size.

To do so, I tried to override the loader and configure it according to our needs:

```js
require('!url-loader?limit=1!../favicon.png')
```

The image will now be referenced by an URL instead of a data uri, however the asset path is missing from the url.

So I tried to set the name option as well.  in https://github.com/xing/hops/blob/d8b1ff3b0e4fb6431b43872d79a54600928964c9/packages/build-config/sections/module-rules/url.js#L12, we set the name of the file to include the asset path.

So naively, I tried something like:

```js
require(`!url-loader?limit=1&name=${hopsConfig.assetPath}/[name]-[hash:16].[ext]!../favicon.png`)
```
Unfortunately, I ended up with an error: 
```
Error: Cannot find module "."
```

After some fiddling around, I came to the conclusion that you simply can not put variables in the options string in `require` calls. (which kind of makes sense, the longer I think about it.)

I took a step back and thought about why are we even using `name` to set the `assetPath` in the first place.
If I recall correctly, it was done to allow html files being put into `/build` while assets should go into `/build/assets` and in earlier versions the html generation also was part of a webpack build which had to respect the `output.path`.

Those days are long gone and we simply write the html files into the `buildDir` on our own.
https://github.com/xing/hops/blob/master/packages/build/lib/generate.js#L21-L38

So, I thought we might be able to get rid off setting the assetPath in loader configs and use the "normal" webpack places instead.

This would allow users to override loaders, but still have their assets put into the correct folder.

This PR is meant as a discussion starter, I'm aware that the code could use some refactoring. 

@ZauberNerd @dmbch does anything I just wrote (and the code in this PR) make any sense to you?

### DIFF
--- a/packages/build-config/configs/build.js
+++ b/packages/build-config/configs/build.js
@@ -10,16 +10,15 @@ var WriteFilePlugin = require('../plugins/write-file');
 var ServiceWorkerPlugin = require('../plugins/service-worker');
 
 var hopsConfig = require('hops-config');
-
-var getAssetPath = path.join.bind(path, hopsConfig.assetPath);
+var publicPath = ('/' + hopsConfig.assetPath + '/').replace(/\/\//, '/');
 
 module.exports = {
   entry: require.resolve('../shims/build'),
   output: {
-    path: hopsConfig.buildDir,
-    publicPath: '/',
-    filename: getAssetPath('[name]-[chunkhash:16].js'),
-    chunkFilename: getAssetPath('chunk-[id]-[chunkhash:16].js'),
+    path: path.join(hopsConfig.buildDir, hopsConfig.assetPath),
+    publicPath: publicPath,
+    filename: '[name]-[chunkhash:16].js',
+    chunkFilename: 'chunk-[id]-[chunkhash:16].js',
   },
   context: hopsConfig.appDir,
   resolve: require('../sections/resolve')('build'),
@@ -52,7 +51,7 @@ module.exports = {
     }),
     new webpack.HashedModuleIdsPlugin(),
     new ExtractTextPlugin({
-      filename: getAssetPath('[name]-[contenthash:16].css'),
+      filename: '[name]-[contenthash:16].css',
       allChunks: true,
       ignoreOrder: true,
     }),
--- a/packages/build-config/configs/develop.js
+++ b/packages/build-config/configs/develop.js
@@ -7,8 +7,7 @@ var webpack = require('webpack');
 var ServiceWorkerPlugin = require('../plugins/service-worker');
 
 var hopsConfig = require('hops-config');
-
-var getAssetPath = path.join.bind(path, hopsConfig.assetPath);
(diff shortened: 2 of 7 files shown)


dmbch:
Of course what you write makes lots of sense. That being said, I am afraid I have to oppose this specific approach, as I feel its potential for subtle breakage (especially downstream) is not really justified by the effect it provides.

Generally, I would not exactly encourage anyone to configure build details inline within their runtime code, except as a (temporary) workaround. Using `module.rules` is also recommended in the web pack [docs](https://webpack.js.org/concepts/loaders/).

Also, I would not rule out returning to a Webpack plugin based rendering implementation at some point. I personally still like the idea of being able to leave all bundle related file I/O to Webpack...

Anyway, back to the actual issue: how about enabling users to 'opt out' of using url-loader by supplying an query parameter for their import? Would that accomplish what you have in mind? Theoretically, it would work like so:

```
require('../favicon.png?no-inline')
```

Strictly speaking, this is also mixing build and runtime code, but, in my opinion, it is much cleaner. This kind of behaviour ought to be configurable like so, though I did not test it yet:

```
{
  test: /\.(png|gif|jpe?g|webp)$/,
  resourceQuery: /^((?!no-inline).)*$/,
  loader: require.resolve('url-loader'),
  options: {
    limit: 10000,
    name: path.join(hopsConfig.assetPath, '[name]-[hash:16].[ext]'),
  },
}
```

p.s.: I afraid I do not quite seem to get why you are using url-loader in the first place - according to its [docs](https://github.com/webpack-contrib/url-loader#limit), what you do (`?limit=1`) seems to force it to fall back to file-loader (albeit an un-configured one).

robin-drexler:
I share the feeling that adjusting the paths the way I did, is rather brittle.

I wasn't aware `resourceQuery` can be used to rule a loader out. And I'd also prefer that approach. It's much simpler and pretty explicit, too.

If I understand the [docs](https://webpack.js.org/configuration/module/#rule-oneof
) correctly, we'd need to have something like this, right? (also not tested yet)

```js
{
  test: /\.(png|gif|jpe?g|webp)$/,
  oneOf: [
    {
      resourceQuery: /no-inline/,
      use: 'url-loader',
      options: {},
    },
    {
      use: 'file-loader',
      options: {},
    },
  ],
};
```

Otherwise no loader would handle the file, because the file-loader is not configured for image files in our case: https://github.com/xing/hops/blob/d8b1ff3b0e4fb6431b43872d79a54600928964c9/packages/build-config/sections/module-rules/file.js#L8

dmbch:
@robin-drexler you are absolutely right: I was thinking of webpack4 branch - there, we configured file-loader as the fallback for all others...

dmbch:
Let us close this for now, then, until we have a better(TM) solution.

PR #439, opened 2018-03-21
fix(build): exit with exitCode 1 when build fails

robin-drexler:
Bug noticed earlier today by a co-worker (can't find his github handle :(), will attribute later.

If `hops-build` fails, it still exits with exitCode 0.

After this PR, `hops-build` will exit with exitCode 1 for webpack errors, build errors and build warnings. I guess it is fair to fail on warnings in a production build.

https://github.com/webpack/docs/wiki/node.js-api#error-handling

### DIFF
--- a/packages/build/commands/build.js
+++ b/packages/build/commands/build.js
@@ -36,7 +36,25 @@ module.exports = function defineBuildCommand(args) {
       if (argv.static) {
         process.env.HOPS_MODE = 'static';
       }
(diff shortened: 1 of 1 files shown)


ZauberNerd:
This breaks for the webpack performance warnings.
I'd propose to fail only on errors - not on warnings.
/cc @dmbch @robin-drexler

PR #441, opened 2018-03-26
wip: use mini-extract-text-plugin instead of ETWP

ZauberNerd:
extract-text-webpack-plugin will not support webpack 4 and is in
maintenance mode.
recommended alternative is mini-extract-text-plugin:
webpack-contrib/extract-text-webpack-plugin#731 (comment)


ZauberNerd:
Apparently the mini-css-extract-plugin has bugs with the watch mode and does not emit a changed file:
https://github.com/webpack-contrib/mini-css-extract-plugin/issues/23
/cc @dmbch @HerrBertling

ZauberNerd:
But since we are not using the plugin in development mode but instead are using the `style-loader` this will not impact our development flow.

PR #445, opened 2018-03-27
fix(hops-build): raise node-mocks-http version to ensure res.locals being set

robin-drexler:
## Current state

Builds can break, because `res.locals` does not exist due to https://github.com/howardabrams/node-mocks-http/commit/baa05e552a5f5414d986b9f7c0a3abc5830d7235

## Changes introduced here
Raise node-mocks-http version to version that ensures res.locals is set

## Checklist


ZauberNerd:
Published with 10.2.1

PR #449, opened 2018-03-29
choose next free port

ZauberNerd:
- hops-config reads port from `process.env.PORT` and defaults to `undefined`
- hops-express reads port from hops-config:
  - if port is undefined it starts at 8080 and tries to find the next free port
  - if port has a value it uses that port and does not try to find a free port

### DIFF
--- a/packages/config/index.js
+++ b/packages/config/index.js
@@ -25,7 +25,7 @@ function applyDefaultConfig(config) {
     {
       https: false,
       host: '0.0.0.0',
-      port: 8080,
+      port: parseInt(process.env.PORT, 10) || undefined,
       locations: [],
       basePath: '',
       assetPath: '',
--- a/packages/express/lib/utils.js
+++ b/packages/express/lib/utils.js
@@ -2,6 +2,7 @@
 
 var fs = require('fs');
 var url = require('url');
+var net = require('net');
 var path = require('path');
 var http = require('http');
 var https = require('https');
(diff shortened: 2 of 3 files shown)


ZauberNerd:
Alternative approach:
- hops-config reads port from `process.env.PORT` and defaults to `undefined`
- hops-express reads port from hops-config:
  - if port is undefined it starts at 8080 and tries to find the next free port
  - if port has a value it uses that port and does not try to find a free port

ZauberNerd:
We might also use this package: https://github.com/sindresorhus/get-port which accepts a preferred port and if that isn't available it lets the OS assign a random unused port.

> f port is omitted or is 0, the operating system will assign an arbitrary unused port, which can be retrieved by using server.address().port
https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_server_listen_port_host_backlog_callback

packages/config/index.js
```diff
@@ -25,7 +25,7 @@ function applyDefaultConfig(config) {
     {
       https: false,
       host: '0.0.0.0',
-      port: 8080,
+      port: parseInt(process.env.PORT, 10) || undefined,
```
dmbch: 👍

packages/express/lib/utils.js
```diff
@@ -19,17 +20,37 @@ function getStatsFromFile() {
   return stats || {};
 }
 
-function defaultCallback(error) {
+function getPort(host, port) {
```
dmbch: It might make sense to limit the number of ports being tried, i.e. servers being started..:

```javascript
const findPort = (ip, port, max) =>
  new Promise((resolve, reject) => {
    max = max || Math.min(65535, port + 50);
    if (port > max) {
      reject(new Error('unable to find free port'));
    } else {
      const server = createNetServer().unref();
      server.on('error', () => resolve(findPort(ip, port + 1, max)));
      server.listen(port, ip, () => server.close(() => resolve(port)));
    }
  });

const getPort = (ip, port) =>
  findPort(ip, ...(Array.isArray(port) ? port : [port || 8080, port]));
```

https://github.com/untool/untool/blob/master/packages/express/lib/run.js#L24-L37

PR #453, opened 2018-04-05
chore: make internal peer dependencies actual dependencies

ZauberNerd:
This makes it consistent with the other packages we have that have internal peer dependencies to other Hops packages. e.g. https://github.com/xing/hops/blob/master/packages/redux/package.json#L22


ZauberNerd:
Closed in favor of #454

PR #454, opened 2018-04-05
chore: set correct dependencies/peerDependencies

ZauberNerd:
This PR marks hops-react and hops-express as peerDependencies with a
range instead of a single version in all packages that depend upon it.

This is only partially correct as we should declare hops-config and
hops-build-config also as peerDependencies in all packages that depend
on them - but that would be a breaking change as these packages are
currently specified as direct dependencies and it would require changing
the templates and all projects that are currently using hops.

### DIFF
--- a/packages/build/package.json
+++ b/packages/build/package.json
@@ -39,7 +39,6 @@
     "filesize": "^3.5.11",
     "hops-build-config": "10.3.0-rc.2",
     "hops-config": "10.3.0-rc.0",
-    "hops-express": "10.3.0-rc.0",
     "memory-fs": "^0.4.1",
     "mkdirp": "^0.5.1",
     "node-mocks-http": "^1.6.7",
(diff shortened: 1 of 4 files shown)


packages/build/package.json
```diff
@@ -39,7 +39,6 @@
     "filesize": "^3.5.11",
     "hops-build-config": "10.3.0-rc.2",
     "hops-config": "10.3.0-rc.0",
-    "hops-express": "10.3.0-rc.0",
```
ZauberNerd: This will be a breaking change - also *technically* all other changes in the peerDependency ranges are breaking.
So therefore we could also move hops-config and hops-build-config to peerDependencies and release this as hops@11.
/cc @dmbch

ZauberNerd:
Closing this until we can find a sensible solution.
Discussion continues in #455

PR #474, opened 2018-04-23
Revert "refactor(react): load scripts async"

ZauberNerd:
This reverts commit 678f2f4f5cec0276d3e6bf822f601662b8b6d8c1.


robin-drexler:
Maybe in hops 11, we can move the scripts to the head and load them with `defer`. 
Scripts would still be executed at the same time as they are now, however in large html files their existence would be noticed faster by the browser, leading to the download starting earlier.

https://hacks.mozilla.org/2017/09/building-the-dom-faster-speculative-parsing-async-defer-and-preload/ "defer and async" is somewhere in the middle. Unfortunately I can not link to it.

It could be considered a breaking change as the defer implementation in older IEs (<10) seems to be buggy. https://caniuse.com/#search=defer

https://hacks.mozilla.org/2017/09/building-the-dom-faster-speculative-parsing-async-defer-and-preload/

PR #490, opened 2018-05-22
Hops next

ZauberNerd:


packages/template-react/src/home/index.js
```diff
@@ -8,7 +8,7 @@ export default function Home() {
       <Helmet>
         <title>Hops Demo</title>
       </Helmet>
-      <h1 className={styles.headline}>Hello World!</h1>
+      <h1 className={styles.headline}>Hello World! {process.env.FOO_ROBIN}</h1>
```
dmbch: I do not really like the idea of using these global variables directly. Instead I would suggest making them available using a HOC.

packages/react/mixin.core.js
```diff
@@ -7,6 +7,12 @@ class ReactCoreMixin extends Mixin {
       require.resolve('babel-plugin-transform-object-rest-spread')
     );
 
+    // TODO: this does not actually work because the webpack browser process
+    // shim also includes other properties, such as `nextTick`, `cwd` and such
+    webpackConfig.node = {
```
dmbch: IMO we should generally not shadow existing globals

packages/template-react/src/home/index.js
ZauberNerd: The issue with using a HoC is that you can't really access environment variables in the mixin methods of a runtime mixin. At least not when it is being executed on the browser.
So if you need access to environment variables in, for example, your `bootstrap()` method, you're out of luck when using a HoC.
We had the idea of setting `global._hopsEnvironmentVariables` in the core mixin and passing that into `enhanceData`, so that you could access environment variables in your mixins through `global._hopsEnvironmentVariables`, but that felt unintuitive.
That's how we arrived at the idea of using webpack to alias the `process.env` namespace. Therefore you could access environment variables in all classes of mixins in the same way.

dmbch: How about introducing a HopsMixin base class that makes those additional variables available as instance props on its constructor?

packages/spec/integration/pwa/__tests__/build.js
```diff
@@ -0,0 +1,37 @@
...
+    jest.setTimeout(30000);
+    await HopsCLI.build();
+    url = await HopsCLI.serve();
+  });
+
+  it('registers a service worker', async () => {
+    const { page } = await createPage();
+    await page.goto(url, { waitUntil: 'networkidle2' });
+
+    // wait until service worker is installed
```
robin-drexler: Wonder if it could make sense to check that the resources are *not* served from the service worker at the first request to exclude potential false positives.

PR #498, opened 2018-06-05
fix: update dependency entries

jhiode:
## Current state

Some dependencies are missing or outdated.

## Changes introduced here

Updated entries in package.json

## Checklist


ZauberNerd:
I believe this PR can be closed as the next branch already contains changes to address these issues.

PR #499, opened 2018-06-05
add Create hops app package

robin-drexler:
Add package to allow using `yarn create hops-app` to initialize a new hops application.

I moved the code from `hops` to this new package and would remove it from `hops` in an additional commit. Leaving the `hops` package as a local cli only.

Is that a direction you'd support @ZauberNerd @dmbch ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,13 @@ $ yarn global add hops
 # OR npm install --global hops
 ```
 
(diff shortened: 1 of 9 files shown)


dmbch:
I really like the general direction! I would personally prefer to keep the (simplified) global mode around, though. We could even consider modifying it so that it offers to shell out to `yarn create` or `npx` if called outside a hops project.

Additionally, I am wondering if we can possibly get rid of the `--npm` argument: if called using `yarn`, we use yarn; if called using `npx`, we go with `npm`.

PR #509, opened 2018-06-28
feat(styled-components): introduce package

ZauberNerd:


ZauberNerd:
@robin-drexler I'd leave it in there just for convenience. It doesn't break or anything if no theme is passed.

PR #511, opened 2018-06-28
feat(typescript): introduce package

ZauberNerd:


packages/typescript/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "preset.js"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/xing/hops.git"
+  },
+  "dependencies": {
+    "@untool/core": "^0.10.0",
+    "ts-loader": "^4.4.2",
+    "typescript": "^2.9.2"
```
ZauberNerd: I'm currently unsure whether we should add typescript as a dependency or peerDependency.
Having it as a dependency adds convenience that users just need to add the "hops-typescript" plugin and everything works.
On the other hand, having it as a peerDependency means that the user needs to install it themselves - but also allows them to choose a different version if they need it (and it reduces peerDependency warnings when using other packages that expect typescript to be installed - eg. tslint, etc).

PR #522, opened 2018-07-18
feat(development-proxy): add development proxy

robin-drexler:
Add a simple proxy for API requests which runs on the same port as the app, mitigating issues with CORS.

Inspired by cra proxy.
Configuration is meant to be simple and limited. For more sophisticated use cases it is probably more feasible to build a custom mixin.

## Configuration

Can be configured in two ways in preset config.
`proxy` string: In this case all<sup>[1]</sup>requests will be proxied to the specified proxy URL.
[1] All meaning every request that is not handled by the webpack dev middleware and that is not a requested html document. Those are handled by the app itself.

```json
{
  "hops": {
    "proxy": "https://example.org"
  }
}
```

or with env var:

```json
{
  "hops": {
    "proxy": "[PROXY_URL]"
  }
}
```

`proxy` can also be an object.

```json
{
  "hops": {
    "proxy": {
      "/foo": { "target": "https://example.org" },
      "/get": { "target": "https://httpbin.org" }
    }
  }
}
```

In that case for each entry a proxy middleware will be created and the config object will be passed to it.

## open questions

- Do we need/want something like that at all?
- Does it make sense to create a dedicated package or should it be part of another preset?

## todo

- [x] tests
- [x] docs
- [x] early return on empty config

### DIFF
--- a/packages/development-proxy/README.md
+++ b/packages/development-proxy/README.md
@@ -1,0 +1,3 @@
+# `hops-development-proxy`
+
+TBD
--- a/packages/development-proxy/README.md
+++ b/packages/development-proxy/README.md
@@ -1,3 +1,88 @@
 # `hops-development-proxy`
 
-TBD
+[![npm](https://img.shields.io/npm/v/hops-development-proxy.svg)](https://www.npmjs.com/package/hops-development-proxy)
+
+[Hops preset](https://missing-link-explain-what-are-presets) that adds an API-proxy for development.
+
+Hops apps are often run on the same host as their backend/api endpoints. However, when working locally, backend apps often run on a different host as the frontend application. This can cause issues with cross origin requests during development that are of no concern in the production environment.
+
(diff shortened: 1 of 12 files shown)


robin-drexler:
@dmbch @ZauberNerd @jhiode what do you think?

dmbch:
LGTM - I am rather convinced we want to have this here. I am not completely sure we want to support the simple, string based config mechanism because it breaks config extension/inheritance.

robin-drexler:
@dmbch Could you elaborate a bit on your concerns regarding the string config version? I don't quite understand them yet, sorry.

dmbch:
@robin-drexler say you have a preset that defines a proxy rule using an object literal. In this case, you will not be able to use string syntax in your project.

If you have a preset that uses string syntax for a proxy definition, you will only be able to override it, but obviously not extend it.

So, if you want to really use config extension/inheritance features for proxy setup, you will always want to use object syntax. Why offer the string version at all, then?

PR #523, opened 2018-07-19
add hooks to allow overriding of data fetching

robin-drexler:
For some use cases, we need to be able to inject a request-aware fetch implementation.

For redux, we'd like to be able to override the thunk middleware to use [extraArguments](https://github.com/reduxjs/redux-thunk#injecting-a-custom-argument).

In Apollo, we'll inject a custom fetch into the `ApolloLink`.

This PR adds hook that'll enable such efforts.

### DIFF
--- a/packages/redux/README.md
+++ b/packages/redux/README.md
@@ -114,3 +114,9 @@ export default render(<MyApp />, {
 #### `getReduxStore(): Store` ([override](https://github.com/untool/mixinable/blob/master/README.md#defineoverride))
 
 Use this method in your own mixins to get a reference to the currently used Redux [Store](https://redux.js.org/api-reference/store) instance.
+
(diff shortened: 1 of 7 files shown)


packages/redux/README.md
```diff
@@ -114,3 +114,9 @@ export default render(<MyApp />, {
 #### `getReduxStore(): Store` ([override](https://github.com/untool/mixinable/blob/master/README.md#defineoverride))
 
 Use this method in your own mixins to get a reference to the currently used Redux [Store](https://redux.js.org/api-reference/store) instance.
+
+#### `getReduxMiddleware(): [middleware]` ([override](https://github.com/untool/mixinable/blob/master/README.md#defineoverride))
```
jhiode: Typo: should be `getReduxMiddlewares`

PR #532, opened 2018-07-26
[RFC]: WIP: prefetch data

ZauberNerd:


packages/graphql/mixin.server.js
```diff
@@ -21,12 +21,19 @@ class GraphQLMixin extends Mixin {
     super(config, element);
 
     this.client = this.createClient(options);
-    this.shouldPrefetchOnServer = undefined;
+    this.prefetchOnServer =
```
ZauberNerd: This also needs to happen in the `hops-redux` package.

packages/graphql/mixin.server.js
```diff
@@ -21,12 +21,19 @@ class GraphQLMixin extends Mixin {
...
 
   bootstrap(request, response) {
-    if (response && response.locals) {
-      this.shouldPrefetchOnServer = response.locals.shouldPrefetchOnServer;
+    if (
+      response &&
+      response.locals &&
+      typeof response.locals.prefetchOnServer === 'boolean'
+    ) {
+      this.prefetchOnServer = response.locals.prefetchOnServer;
```
ZauberNerd: Only when `response.locals.prefetchOnServer` is actually set will we override the current behavior from the config.

packages/graphql/mixin.server.js
ZauberNerd: By default `prefetchOnServer` is `true` and we only do this, so that people can disable it entirely through their config.

packages/graphql/mixin.server.js
ZauberNerd: I suppose we should also add `this.prefetchOnServer !== false`, so that users actually **can** disable prefetching via config.

PR #542, opened 2018-07-31
chore: upgrade dependencies and recreate yarn.lock

ZauberNerd:


packages/webpack/mixin.core.js
```diff
@@ -18,6 +18,9 @@ class LoggerPlugin {
 class WebpackCoreMixin extends Mixin {
   configureBuild(webpackConfig) {
     webpackConfig.plugins.push(new LoggerPlugin());
+    webpackConfig.resolve.mainFields = webpackConfig.resolve.mainFields.filter(
```
jhiode: What's the reason for this change?

PR #554, opened 2018-08-06
fix: upgrade untool

robin-drexler:


robin-drexler:
closed in favor of https://github.com/xing/hops/pull/545

PR #561, opened 2018-08-13
refactor(lambda): use NODE_ENV instead of CLI args to start prod mode

ZauberNerd:

_This is a template. Feel free to remove the parts you do not need! Start with this line, for example_ :wink:

* [ ] All commit messages adhere to the [appropriate format](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit) in order to trigger a correct release
* [ ] All code is written in plain ECMAScript v5 and is formatted using [prettier](https://prettier.io)

### DIFF
--- a/packages/lambda/lambda.js
+++ b/packages/lambda/lambda.js
@@ -1,6 +1,7 @@
 'use strict';
 
 process.env.UNTOOL_NSP = 'hops';
+process.env.NODE_ENV = 'production';
 
 const serverlessHttp = require('serverless-http');
 const config = require('@untool/core/lib/env').environmentalize(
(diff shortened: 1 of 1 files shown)


ZauberNerd:
@robin-drexler with these changes from untool (https://github.com/untool/untool/pull/104) it won't work anymore - we'd have to use
```javascript
const configure = require('@untool/express').configure;
const { createServer } = configure(config, { production: true });
```

robin-drexler:
Gotcha, thanks for explaining. :)

dmbch:
@robin-drexler the previous version was a bit problematic, to say the least: setting `NODE_ENV` to `production` is essential to configure Express for live environment, for example. See: https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production

PR #599, opened 2018-09-03
fix(cli): print getting started message based on availability of CLI

ZauberNerd:
Print contextual getting started message based on availability of Hops
global CLI, yarn or npm.

### DIFF
--- a/packages/cli/lib/init.js
+++ b/packages/cli/lib/init.js
@@ -2,6 +2,7 @@
 
 var fs = require('fs');
 var path = require('path');
+var execSync = require('child_process').execSync;
 var tar = require('tar');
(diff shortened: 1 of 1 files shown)


packages/cli/lib/init.js
```diff
@@ -101,8 +111,20 @@ function init(root, appName, options) {
         console.log('Hooray \\o/');
         console.log('Your project has been successfully created.');
         console.log(
-          'You should change into its directory and execute "hops" to see a list',
-          'of available commands.'
+          'You should change into its directory and execute',
+          isCLIInstalledGlobal()
```
robin-drexler: bike shed: It should be `isCLIInstalledGlobally` as it is an adverb. :D

PR #600, opened 2018-09-03
Upgrade untool babel

robin-drexler:


packages/spec/integration/jest-preset/__tests__/preset.js
```diff
@@ -0,0 +1,6 @@
+describe('jest-preset', () => {
+  it('allows to use typescript', () => {
+    const calculator = require('./setup/calculator.ts').default;
```
ZauberNerd: Do you need to use `.ts` here? Or could you just require `'./setup/calculator` instead?

packages/spec/integration/typescript/package.json
```diff
@@ -9,11 +9,13 @@
     "testEnvironment": "../../helpers/env.js"
   },
   "scripts": {
-    "start": "node -e 'require(\"hops\").run(\"start\")'"
+    "start": "node -e 'require(\"hops\").run(\"start\")'",
+    "tsc": "tsc"
```
ZauberNerd: I think renaming the script to `type-check` makes more sense.

packages/styled-components/package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/xing/hops.git"
   },
   "dependencies": {
-    "babel-plugin-styled-components": "^1.5.1",
+    "babel-plugin-styled-components": "1.5.1",
```
ZauberNerd: Can we use `1.6.1` now? Or should we wait for your PR to get merged and then go to the next version?

robin-drexler: I'd wait as 1.6.1 is still rather slow

packages/spec/integration/typescript/package.json
robin-drexler: Fine with me, will adjust it in Readme as well then

Issue #608, opened 2018-09-04
hops-lambda: Printing wrong link after deploy


omelhoro:
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

I have the version 11.0.0-rc.31 of the lambda plugin and following hops config:
```
  "hops": {
    "browsers": "last 1 Chrome versions",
    "locations": [
      "/",
      "/counter"
    ],
    "assetPath": "prod",
    "basePath": "prod",
    "node": "8.10"
  },
```
When I deploy per `hops lambda deploy` I get a link which ends with `.amazonaws.com/rod/` so when I click on it AWS says 'access forbidden', it only works when I change the end to `/prod/`.

PR #609, opened 2018-09-04
Fix hops-lambda URL

ZauberNerd:
This pull request closes issue #608

### DIFF
--- a/packages/lambda/lib/aws-config.js
+++ b/packages/lambda/lib/aws-config.js
@@ -1,6 +1,12 @@
 'use strict';
 
 const semver = require('semver');
+const {
+  stripLeadingSlash,
+  stripTrailingSlash,
+} = require('@untool/express').uri;
+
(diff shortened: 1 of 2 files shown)


packages/lambda/lib/aws-config.js
```diff
@@ -1,6 +1,12 @@
 'use strict';
 
 const semver = require('semver');
+const {
+  stripLeadingSlash,
+  stripTrailingSlash,
+} = require('@untool/express').uri;
+
+const trimSlashes = input => stripLeadingSlash(stripTrailingSlash(input));
```
robin-drexler: Why not put `trimSlashes` into a file inside /lib and use it here and in `lambda.js`?

ZauberNerd: I created a PR in untool: https://github.com/untool/untool/pull/125

PR #613, opened 2018-09-05
fix(typescript): install necessary plugins and presets for typescript

ZauberNerd:
In order to use TypeScript with the new @babel/preset-typescript we need
to remove the flow-preset and add / change the decorators and
class-properties plugins.

With this change stripping types from .tsx? files mostly works.
The only issues that I have found are:
- https://github.com/babel/babel/issues/7749
- https://github.com/babel/babel/issues/8361

### DIFF
--- a/packages/typescript/mixin.core.js
+++ b/packages/typescript/mixin.core.js
@@ -1,12 +1,53 @@
 const { Mixin } = require('hops-mixin');
 
+const getPluginDetails = plugin => {
+  return typeof plugin === 'string' ? [plugin] : plugin;
+};
+
+const findPlugin = needle => {
+  return plugin => {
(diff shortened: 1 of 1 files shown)


ZauberNerd:
After discussing with @robin-drexler we decided to switch back to using `ts-loader` and `typescript` as we did before in this preset.

Because with the Babel variant there are two kind of important features missing (type imports and interface exports / re-exports) - at least that we know of now.

Other reasons for reverting:
- The babel variant will probably always be running a bit behind in terms of new features
- chasing bugs between babel / typescript etc and not knowing exactly where they come from
- using ts-loader and typescript works and there is no real benefit to switching to babel (at least for now)

Issue #619, opened 2018-09-07
Can't override getApolloLink


tjwelde:
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

I want to override `getApolloLink` from `hops-graphql`.
I am doing it like this:

```
const { Mixin } = require('hops-mixin');
const { HttpLink } = require('apollo-link-http');

class GraphqlMixin extends Mixin {
  getApolloLink() {
    [...]
    return link;
  }
}

module.exports = GraphqlMixin;
```

As filename, I tried it with mixin.{server, browser, runtime}.js  
Other mixins in the same directory work.

## Proposed solution

My guess is, that my mixin is somehow prioritised less than the `hops-graphql` library, or I am doing something wrong.

ZauberNerd:
@tjwelde I just tried to reproduce this and can confirm, that the mixin implementation is not being called.
I will investigate further.. :male_detective:

PR #625, opened 2018-09-10
Fix GraphQL mixin and fragment type caching

ZauberNerd:
This pull request closes issue #619

We had a bug in our GraphQL mixin that prevented us from overriding the
exposed `getApolloLink` mixin hook.

The problem was that we called mixin methods inside the constructor, but
at that time the mixin container has not been created, so when some
other mixin tries to override the mixin method it will never be called.

Also while working on this problem I found out that our introspection
results are never actually used, because a new instance is created for
each request and therefore `this.introspectionResult` was always set to
`undefined`.

I moved the instantiation of the apollo client from the constructor to
the `bootstrap` method in order to allow for `getApolloLink` to use an
overridden implementation.

I also moved the fragment types reading / caching mechanism to the
constructor and the cache itself outside of the class as a module-scoped
variable so that the generated fragment types are being used on both
server- and client-side.

### DIFF
--- a/packages/graphql/mixin.browser.js
+++ b/packages/graphql/mixin.browser.js
@@ -19,7 +19,11 @@ class GraphQLMixin extends Mixin {
   constructor(config, element, { graphql: options = {} } = {}) {
     super(config, element);
 
-    this.client = this.createClient(options);
+    this.options = options;
+  }
+
+  bootstrap() {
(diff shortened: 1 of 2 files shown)


tjwelde:
Awesome! I was wondering about the introspection, too. Nice to see it solved already!

PR #632, opened 2018-09-12
Rewrite CLI to use mixins for logging purposes

ZauberNerd:
At the moment the CLI does not use a mixin and log output for build and
server is done in the hops-express and hops-webpack package.

Inspired by the CLI package in untool, we now use a CLI mixin which
takes care of logging for express and webpack.

~~**TODO:** With this PR the webpack readme will be removed - we should
make sure that the necessary information is retained someplace else.~~

Edit: The express and webpack readmes are now part of: https://github.com/xing/hops/pull/655


packages/cli/mixin.core.js
```diff
@@ -0,0 +1,83 @@
...
+          'Compiled successfully',
+          'in',
+          prettyMS(stats.endTime - stats.startTime)
+        );
+      }
+    });
+  }
+}
+
+module.exports = class CLIMixin {
```
jhiode: We could add `extends Mixin` here to make it clear(er) it's a mixin.

That would also be good in order to not break as soon as it is actually required to inherit from Mixin.
And thats what we are doing anywhere else.

ZauberNerd: I don't think we can - at least not while the mixin is still part of `@untool/core`, because then the CLI (which might be installed globally) will have a dependency to `@untool/core`.

But what we could do, is to make the `Mixin` a separate package in untool and then we could add it (or rather `hops-mixin`) as a dependency to the CLI package.

/cc @dmbch

dmbch: As the `Mixin` base class is just a five-liner, I am opposed to releasing it separately. I personally think that for this very use-case, it is completely fine to have a standalone class.

PR #642, opened 2018-09-18
GraphQL mock server: Compile executable mock schema through webpack

ZauberNerd:
In order to allow writing mock schemas in ES6 and beyond (and to allow importing `*.graphql` files via a webpack loader) we now compile the mock schema via webpack.

This is a breaking change, as it is now required to export an executable schema instead of configuration which the mock server would then turn into an executable schema.

This change is intended to reduce complexity in our GraphQL setup and allow for greater flexibility in schema stitching because now the application developer needs to do the stitching themselves and provide the executable schema as entry point for the mock server.


packages/graphql/mock-server/mixin.core.js
```diff
@@ -1,232 +1,30 @@
...
-const { print } = require('graphql');
-const {
-  makeRemoteExecutableSchema,
-  makeExecutableSchema,
-  mergeSchemas,
-  addMockFunctionsToSchema,
-  introspectSchema,
-} = require('graphql-tools');
-const { sync: resolve } = require('enhanced-resolve');
+const renderMiddleware = require('@untool/webpack/lib/middleware/render');
```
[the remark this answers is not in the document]
ZauberNerd: Not sure, as this is the first use-case of the render middleware outside of untool. Maybe once we have one more use-case for it?

packages/graphql/mock-server/mixin.core.js
```diff
@@ -1,232 +1,30 @@
...
-        console.warn(`Skip: can't use mocks from ${path}`, e);
-      }
-    });
-  }
-
-  configureServer(rootApp, middlewares) {
     if (process.env.NODE_ENV === 'production') {
       console.warn(
-        'Skip: GraphQL Mock-Server is not supported in production mode.'
+        'It is not recommended to run the GraphQL Mock Server in production.'
```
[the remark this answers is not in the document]
ZauberNerd: We decided against that, because we want to be able to start a mock server in production mode, but print a warning if that happens.

packages/graphql/README.md
```diff
@@ -169,3 +185,34 @@ This is an overrideable hook that can be used to customize the behavior of when
...
+
+**Supports Local GraphQL Playground against your GraphQL schema**
+
+`open http://localhost:<port>/graphql`
+
+![GraphiQL Playground](./playground.png)
+
+## GraphQL Mock-Server usage examples
+
+- [Custom field resolver](../../spec/integration/graphql-mock-server/mocks/exercise1)
```
ZauberNerd: These links are broken.

packages/graphql/lib/mock-server-middleware.js
```diff
@@ -0,0 +1,100 @@
...
+  mergeSchemas,
+} = require('graphql-tools');
+const fetch = require('isomorphic-fetch');
+
+/* eslint-disable node/no-missing-require */
+const {
+  schemas = [],
+  resolvers = [],
+  mocks = {},
+} = require('hops-graphql-mocks');
```
ZauberNerd: Add a comment to describe that this is an alias to the `graphqlMocks` preset config option.

packages/graphql/mixin.browser.js
```diff
@@ -14,6 +14,7 @@ const {
   IntrospectionFragmentMatcher,
   HeuristicFragmentMatcher,
 } = require('apollo-cache-inmemory');
+const introspectionQueryResultData = require('./fragment-types.json');
```
ZauberNerd: Add a comment describing that this is handled by the `fragment-types-loader.js`

packages/graphql/mixin.core.js
```diff
@@ -1,55 +1,69 @@
...
+    const webpackConfig = this.getBuildConfig('node');
+
+    webpackConfig.output.filename = 'hops-graphql-mock-server.js';
+
+    Object.assign(webpackConfig.resolve.alias, {
+      'hops-graphql-mocks': require.resolve(this.config.graphqlMocks),
+      '@untool/entrypoint': require.resolve('./lib/mock-server-middleware'),
+    });
+
+    const configLoaderIndex = webpackConfig.module.rules.findIndex(
```
ZauberNerd: Add a comment explaining why we remove this plugin.

packages/graphql/mixin.server.js
```diff
@@ -16,33 +15,12 @@ const {
   IntrospectionFragmentMatcher,
   HeuristicFragmentMatcher,
 } = require('apollo-cache-inmemory');
-
-let introspectionResult = undefined;
-let warned = false;
+const introspectionQueryResultData = require('./fragment-types.json');
```
ZauberNerd: Add a comment describing that this is handled by the `fragment-types-loader.js`

packages/graphql/preset.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
-  fragmentsFile: '<rootDir>/fragmentTypes.json',
   graphqlSchemaFile: '',
   graphqlUri: '',
-  mixins: [__dirname, './mock-server'],
+  enableGraphqlMockServer: process.env.NODE_ENV !== 'production',
```
ZauberNerd: `graphqlMocks` is missing.

packages/graphql/mixin.core.js
```diff
@@ -1,55 +1,69 @@
...
-                console.error('Could not fetch GraphQL fragments:');
-                console.trace(err);
-              });
-          },
-        })
-        .help('help')
-        .alias('h', 'help')
-        .demandCommand()
+  configureServer(rootApp, middleware) {
+    if (!this.config.graphqlMocks || !this.config.enableGraphqlMockServer) {
```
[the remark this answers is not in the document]
ZauberNerd: What do you propose we do, when we render static pages? In my opinion the Apollo Server should start for static rendering, so that the rendered pages contain all the data, right?
Controlling if/how the pages should be rendered can be done via `shouldPrefetchOnServer`.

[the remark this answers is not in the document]
jhiode: IIRC `static` is set when running `hops build`, when running `yarn serve` the mode is `serve`.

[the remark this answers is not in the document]
ZauberNerd: No, that's not entirely correct. The mode argument can either be `develop`, `serve` or `static`.
- `develop` means that the server has been started in development mode `hops develop` or `hops start` and in that case we want to start the apollo mock server.
- `serve` means that the server has been started in serving mode - started via `hops serve` or `hops start -p`.
- `static` is only there for static rendering and will only occur when a static build is started: `hops build -s` or `hops start -ps`.

So, I'd say the configuration that makes the most sense would be to only run the apollo server is when `mode === develop || mode === static` for now. And for `mode === serve` we should try to find another solution. For example we could use `configureBuild` to create a production ready apollo mock server (similar to how we are building the `server.js` middleware for production builds).

packages/graphql/README.md
```diff
@@ -43,20 +39,53 @@ This argument can be specified multiple times to add multiple HTTP headers.
...
+| Name                      | Type      | Default                                 | Required | Description                                                              |
+| ------------------------- | --------- | --------------------------------------- | -------- | ------------------------------------------------------------------------ |
+| `fragmentsFile`           | `String`  | `<rootDir>/fragmentTypes.json`          | _no_     | Where to store the generated fragment types file                         |
+| `graphqlSchemaFile`       | `String`  | `''`                                    | _no_     | Path to your GraphQL schema file                                         |
+| `graphqlUri`              | `String`  | `''`                                    | _yes_    | Url to your GraphQL endpoint                                             |
+| `shouldPrefetchOnServer`  | `Boolean` | `true`                                  | _no_     | Whether Hops should execute GraphQL queries during server-side rendering |
+| `enableGraphqlMockServer` | `Boolean` | `process.env.NODE_ENV !== 'production'` | _no_     | Whether Hops should start an Apollo GraphQL Mock Server                  |
```
jhiode: I think this should be an environment variable, the default could still be `process.env.NODE_ENV !== 'production'` though.

We have (mostly / at least) 4 kinds of configurations:

1. True isomorphic configuration or configuration that is needed on the client.
2. Configuration that is only needed on the server but doesn't hurt to be on the client as well.
3. Environment dependent configuration.
4. Sensitive informations like tokens and secrets.

And I would like to use environment variables for type 3 and 4 and in my opinion `enableGraphqlMockServer` is type 3.

PR #662, opened 2018-10-04
fix: correct CORS headers for graphql mock server

matthias-reis:
affects: hops-graphql

## Current state

graphql mock server requests are answered with invalid CORS headers

## Changes introduced here

leverage CORS config options in apollo middleware / server and configure them correctly

## Checklist

### DIFF
--- a/packages/graphql/mock-server/mixin.core.js
+++ b/packages/graphql/mock-server/mixin.core.js
@@ -174,7 +174,10 @@ module.exports = class GraphQLMockServerMixin extends Mixin {
     const app = express();
     app.use(cookieParser());
 
-    const { remoteSchemas = [] } = this.graphqlMockServerConfig;
+    const {
+      remoteSchemas = [],
+      middlewareConfig = {},
+    } = this.graphqlMockServerConfig;
 
     const remoteSchemaOptions = remoteSchemas.map(value =>
       mapRemoteSchemaOptions(value)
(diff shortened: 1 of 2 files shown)


matthias-reis:
Updated it, but I additionally had to add credentials: include at several places, which I think doesn't need to be configurable ...

packages/graphql/mock-server/mixin.core.js
```diff
@@ -57,6 +57,7 @@ module.exports = class GraphQLMockServerMixin extends Mixin {
 
   async fetcher({ query, variables, operationName, url, headers }) {
     const fetchResult = await fetch(url, {
+      credentials: 'include',
```
robin-drexler: Why have you added `credentials` here? Isn't this executed on the server where there are no cookies?

packages/graphql/mock-server/mixin.core.js
```diff
@@ -210,9 +214,17 @@ module.exports = class GraphQLMockServerMixin extends Mixin {
       const server = new ApolloServer({
         schema: this.schema,
         context: ({ req }) => ({ req, config: this.config }),
+        playground: {
+          settings: {
+            'request.credentials': 'include',
```
robin-drexler: For what exactly is this needed? Which request will then include credentials?

robin-drexler:
Can you maybe add some docs (maybe even the cors example) here? https://github.com/xing/hops/blob/master/packages/graphql/mock-server/README.md

packages/graphql/mock-server/mixin.core.js
matthias-reis: This is for the Request from the playgound (former graphiQL) to theserver.

packages/graphql/mock-server/mixin.core.js
matthias-reis: This has been added by Mike in my node modules folder. But you're right this makes no sense to me either at the moment. It's not needed for the current mock server use case anyway, so I've removed it for now.

PR #676, opened 2018-10-15
test: set jest timeout globally to 5 minutes (for each file)

ZauberNerd:
Our tests are sometimes failing with a jest timeout.

Therefore I have moved the timeout setup to a global
`setupTestFrameworkScriptFile` in which the timeout is set to 5 minutes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,8 +114,8 @@
   ],
   "jest": {
     "projects": [
-      "packages/!(spec)",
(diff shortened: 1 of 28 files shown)


robin-drexler:
graphql tests are still failing, although it seems to be caused by puppeteer taking longer to navigate: https://travis-ci.org/xing/hops/jobs/441744868#L581

PR #677, opened 2018-10-15
fix(styled-components): upgrade styled components

robin-drexler:
https://www.styled-components.com/releases
https://www.styled-components.com/docs/faqs#what-do-i-need-to-do-to-migrate-to-v4

I think we should include this major release before going latest with hops 11.


ZauberNerd:
Closed via #681

Issue #418, opened 2018-02-26
Can not load css for external package


lipina:
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

Following the [documentation](http://react-day-picker.js.org/docs/getting-started) for setting up the `react-day-picker` in my project I try to require styles usying `css-loader` like this: 
```
import "react-day-picker/lib/style.css";
```
but it doesn't work.

## Proposed solution

🤔

dmbch:
Hey @lipina, thanks for this report! I am a afraid it appears that this might just be `hops`‘ current behavior: we are currently using `css-loader` with CSS modules for all CSS files.

As soon as we can merge #417 or #354 we will have a per-file opt-out for this behavior (see bc3a94a)

Could you temporarily include that component’s CSS locally in your project and wrap its class identifiers like so `:global(.xxx)` (see https://github.com/css-modules/css-modules/blob/master/README.md#exceptions)?

dmbch:
This issue can now (finally) be resolved: https://github.com/xing/hops/releases/tag/v10.3.0

@lipina please verify.

lipina:
It doesn't work for me unfortunately. 🤷‍♀️

I tried both `import "react-day-picker/lib/style.css";` and `import "react-day-picker/lib/style.css?global";`

ZauberNerd:
This has been merged into master with #688 and will be included in the next release of Hops 11 RC.

PR #688, opened 2018-11-01
fix(postcss): re-introduce global CSS support

herschel666:
refs #418


ZauberNerd:
I'd argue that this is a `feat` rather than a `fix` ;)

robin-drexler:
@ZauberNerd global css support has been in hops 10 (albeit a bit broken when it came to node_modules), so we're just preventing unnecessary a breaking change here.

PR #691, opened 2018-11-02
test: downgrade yarn to fix never ending jest process

robin-drexler:


ZauberNerd:
@robin-drexler yes, the problem was that we tried spawning yarn to run scripts from the package.json in order to run integration tests - not using yarn, but hops directly solved the issue: https://github.com/xing/hops/commit/826bfa4f310cf96d4d20ec6d0213fe95b5c34b25

robin-drexler:
@ZauberNerd haha we just had the same idea for running the tests. Use jest directly instead of yarn. If it works, I guess that's a viable workaround as well(?)

klyrr:
aha!

robin-drexler:
@klyrr Google? :D

klyrr:
I think I was looking for release notes for yarn 1.12. and somehow stumbled about the issue that you were referring.
And I was thinking ... ha! I know that name!

robin-drexler:
It's a small world. Why don't you check out hops while your here? :)

herschel666:
@klyrr This could be the start of a Yarn self-help group. 😅

PR #700, opened 2018-11-15
chore: upgrade untool and other dependencies

jhiode:
The latest puppeteer package breaks our tests. I've set the version to the latest working version (1.9.0) in the package.json.


ZauberNerd:
@jhiode could you edit the commit message of the first commit and explain why we pin the puppeteer version?

PR #703, opened 2018-11-19
replace shouldPrefetchOnServer with canPrefetchOnServer hook

jhiode:
Instead of asking just one mixin whether it is ok to prefetch or not, we can now ask all relevant mixins and if at least one mixin disagrees we prevent prefetching.

### DIFF
--- a/packages/redux/action-creator-dispatcher/mixin.server.js
+++ b/packages/redux/action-creator-dispatcher/mixin.server.js
@@ -2,7 +2,7 @@
 const { createLocation } = require('history/LocationUtils');
 const {
   strategies: {
-    sync: { override },
+    sync: { sequence },
   },
 } = require('hops-mixin');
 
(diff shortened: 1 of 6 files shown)


packages/redux/action-creator-dispatcher/mixin.server.js
```diff
@@ -11,14 +11,16 @@ const ReduxActionCreatorCommonMixin = require('./mixin.runtime-common');
 class ReduxActionCreatorServerMixin extends ReduxActionCreatorCommonMixin {
   bootstrap(request) {
     this.request = request;
-    this.prefetchedOnServer = this.shouldPrefetchOnServer();
+
+    const prefetchOnServer = this.canPrefetchOnServer().every(value => value);
```
ZauberNerd: Can you make this equivalent to `hops-graphql` and execute `this.canPrefetchOnServer` in the `fetchData()` hook?

PR #711, opened 2018-11-22
remove global CLI and make "hops" the default preset

ZauberNerd:


packages/cli/package.json
```diff
@@ -13,19 +13,32 @@
...
-    "tar": "^4.4.6",
-    "validate-npm-package-name": "^3.0.0",
-    "yargs": "^12.0.1"
+    "hops-express": "11.0.0-rc.49",
+    "hops-mixin": "11.0.0-rc.49",
+    "hops-react": "11.0.0-rc.49",
+    "pretty-ms": "^3.2.0"
+  },
+  "peerDependencies": {
+    "react": "^16.3.0",
```
herschel666: do we want to align the versions with the ones from the `devDependencies`?

packages/cli/core.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+const { Mixin, strategies } = require('hops-mixin');
+
+module.exports = {
+  Mixin,
+  strategies,
```
jhiode: Should we export `hops-config` as `config` here as well?

packages/cli/runtime.js
```diff
@@ -0,0 +1,29 @@
...
+module.exports = {
+  Header,
+  Import,
+  Miss,
+  Status,
+  render,
+  ServerDataContext,
+  withServerData,
+  ConfigContext,
+  withConfig,
```
jhiode: And here as well?

packages/create-hops-app/lib/package-manager.js
```diff
@@ -97,3 +97,10 @@ function isPackageInstalled(name) {
   }
 }
 module.exports.isPackageInstalled = isPackageInstalled;
+
+function isUsingNpxOrNpm(env) {
```
jhiode: What is the reason to pass the env map instead of accessing `process.env` here?

packages/jest-preset/jest-preset.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
       'jest-preset-hops/mocks/file.js',
     '^.+\\.tpl$': 'jest-preset-hops/mocks/tpl.js',
     '^.+\\.css$': 'identity-obj-proxy',
+    hops: 'hops/runtime.js',
```
jhiode: Should we do the same for `hops-react` (and if needed other packages as well)?
We should at least remove[ the deep import](https://github.com/xing/hops/blob/2cd4b1162b375a35e4cc964f41d14438db2e3496/packages/react/index.js#L7) again.

packages/cli/core.js
ZauberNerd: As discussed, we won't include the `hops-config` for now in the default preset.

packages/cli/runtime.js
ZauberNerd: As discussed, we won't include the `hops-config` for now in the default preset.

packages/create-hops-app/lib/package-manager.js
ZauberNerd: It makes the function side-effect free and, IMHO makes it easier to understand [the callsite](https://github.com/xing/hops/pull/711/files#diff-4544895faadfb1612bf6b35db49e970dR34) because you know that this function will return something that is calculated from the env variables.

packages/cli/package.json
ZauberNerd: I've created a separate PR to fix out-of-alignment versions: https://github.com/xing/hops/pull/728

packages/jest-preset/jest-preset.js
ZauberNerd: I would argue against removing the deep import, because `@untool/react`'s `"main"` export is implicitly the [`index.js`](https://github.com/untool/untool/blob/master/packages/react/index.js) and the deep import to `@untool/react/lib/runtime.js` will only be picked up through the `"browser"` and `"server"` fields.
Therefore we won't be able to use the runtime components in jest, because jest does not understand `browser` or `server` (or at least not without configuration - which would complicate other things).

And the `hops-react` package only has a single `"main"` field, wich already contains the runtime components, therefore a jest alias for `hops-react` is not necessary.

ZauberNerd:
Todo: rename `cli` folder to something else..
- `main`
- `dist`
- `hops`
?

jhiode:
As the package has the name `hops` I would vote for just that...

PR #716, opened 2018-11-23
feat(react): add withConfig hoc and context

robin-drexler:


packages/react/index.js
```diff
@@ -16,4 +19,6 @@ module.exports = {
   render,
   ServerDataContextConsumer: Consumer,
   withServerData,
+  ConfigConsumer,
```
ZauberNerd: I'm not too happy that we have `ServerDataContextConsumer` and `ConfigConsumer`, should we rename one of them?
Maybe we can rename `ServerDataContextConsumer` to `ServerDataConsumer` **and** keep `ServerDataContextConsumer`?

robin-drexler: I agree and added a commit for it. https://github.com/xing/hops/commit/e48fe1cd0fb52a961b90cc3e16eb66fe85fb4b71

PR #718, opened 2018-11-26
[RFC]: chore: move integration tests to examples/ folder

ZauberNerd:


examples/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
herschel666: what purpose does this `package.json` serve?

ZauberNerd: As far as I know it is needed because this is also a package in the workspace.

herschel666: only its contents should be packages. `./packages` doesn't have a `package.json`.

ZauberNerd: You are right, we could get away without this `package.json`, but if we want to lint the examples, we need to put the eslint config into our global config, which would make our global config less strict and we also have to ignore some eslint rules..
See this commit: https://github.com/xing/hops/pull/718/commits/ce2e293590e8d9b4c58ba6b4a2809899c6e0b383

herschel666: What about about "outsourcing" the config into `.eslintrc.js` files and have the `/examples` one extend the one in the root of the project?

ZauberNerd:
Closing this for now. Maybe we will revisit this topic later when we have more time.

PR #721, opened 2018-11-27
fix(redux): always set prefetchedOnServer member variable

jhiode:
This fixes a bug that were introduced by #703 

The `prefetchedOnServer` member variable is used in [`packages/redux/action-creator-dispatcher/mixin.runtime-common.js`](https://github.com/xing/hops/blob/1b4b8eaa7c9f5c792ee3ce5ddc5638b4aa7d68a9/packages/redux/action-creator-dispatcher/mixin.runtime-common.js#L84)


jhiode:
Without setting `this.prefetchedOnServer = true;` on the server, the client would never get that information and has to refetch everything again.

jhiode:
I've created an issue for the missing tests: #723

PR #727, opened 2018-12-05
fix: upgrade untool packages

ZauberNerd:


jhiode:
As untool has increased the minimum version of `react` and `react-dom` peer dependenies to `^16.4.0`, we should do the same.

PR #728, opened 2018-12-05
Fix dependency issues (minimum required versions, peer dependencies, etc)

ZauberNerd:
A couple of Hops packages have incorrect dependency versions in their
package.json files and some packages are missing required peer
dependencies.

This PR adds missing peer dependencies and increases the minimum
required versions of other dependencies to remove all peer dependency
warnings in Hops.

### DIFF
--- a/packages/jest-preset/package.json
+++ b/packages/jest-preset/package.json
@@ -27,6 +27,7 @@
     "@babel/plugin-transform-flow-strip-types": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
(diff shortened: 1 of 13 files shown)


packages/jest-preset/package.json
```diff
@@ -27,6 +27,7 @@
     "@babel/plugin-transform-flow-strip-types": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
jhiode: ❤️

PR #741, opened 2018-12-11
feat(development-proxy): add proxy mixin hooks

robin-drexler:


packages/development-proxy/mixin.core.js
```diff
@@ -20,29 +31,32 @@ class ProxyMixin extends Mixin {
...
+          )
         )
       );
     }
 
     if (typeof proxyConfig === 'object') {
       debug('Using proxy object version', proxyConfig);
 
-      Object.entries(proxyConfig).forEach(([path, config]) => {
+      Object.entries(proxyConfig).forEach(([path, { target }]) => {
```
jhiode: I this the only option that we want to support in the config?

robin-drexler: Not sure, but we should not give the impression that someone could use the functions here.

We can always add more later, too.

What would your suggestion be?

PR #743, opened 2018-12-13
fix(jest-preset-hops): Use correct runtime file in moduleNameMapper

robin-drexler:


packages/spec/integration/jest-preset/__tests__/preset.js
```diff
@@ -8,4 +8,9 @@ describe('jest-preset', () => {
     const calculator = require('./setup/calculator-flow').default;
     expect(calculator(2, 40)).toEqual(42);
   });
+
+  it('allows to require hops runtime code', () => {
+    const hops = require('hops');
+    expect(hops.render).not.toBeUndefined();
```
jhiode: What about `expect(hops.render).toBeDefined()`?

robin-drexler: Meh, you're correct. That'd be better. Changed it and am now waiting for travis again. :D

PR #745, opened 2018-12-13
docs: only use npm install examples in readmes

robin-drexler:


robin-drexler:
@herschel666 ooops, cheers! Fixed for `hops-preset-defaults` & `hops-development-proxy`. `hops` will be addressed by another PR.

PR #752, opened 2018-12-17
fix(create-hops-app): remove version tag from npx install success notice

ZauberNerd:


packages/create-hops-app/lib/init.js
```diff
@@ -112,7 +112,11 @@ function init(root, appName, options) {
           'To see a list of available commands through Hops presets execute:',
           pm.hasBeenInstalledViaYarn(appRoot)
             ? '`yarn hops`'
-            : '`npx hops@' + options.hopsVersion + '`'
+            : `\`npx hops${
```
herschel666: Doesn't `npx hops` always resolve to the Hops CLI being located in the local `node_modules/.bin`-folder, since it's shipped with the `hops` base package?

ZauberNerd: I don't think so. At least a quick test doesn't look like it:
- `cd /tmp`
- `npx create-hops-app my-app`
- `cd my-app`
- `npx hops` <-- seems to resolve to the local nstall
- `npx hops@latest` <-- seems to re-download every time you execute this command

herschel666: I was rather concerned about `npx hops@<some version>`. I can't imagine a case where this wouldn't resolve to the local hops, even when it's not the latest version?!

jhiode: Yes, the question is why do we need the `@version` in the first place? And why only for npx?

jhiode: Or to rephrase it: AFAIC it should be fine to do just `npx hops`.

ZauberNerd:
@herschel666 @jhiode please take another look

jhiode:
Approved with all my accounts 😂

ZauberNerd:
Then it must be good!

PR #756, opened 2018-12-20
docs: add jest-preset-hops to main readme

robin-drexler:

PR #760, opened 2018-12-21
stabilize hmr test

robin-drexler:
This tests breaks pretty often on travis, presumably due to resource constraints.
https://travis-ci.org/xing/hops/builds/ 

E.g. I restarted https://github.com/xing/hops/pull/756 five times already and this test always broke (but in different jobs/node versions)

### DIFF
--- a/packages/spec/integration/hot-module-reload/__tests__/develop.js
+++ b/packages/spec/integration/hot-module-reload/__tests__/develop.js
@@ -11,7 +11,10 @@ function changeFile(id) {
   fs.writeFileSync(path.join(global.cwd, 'index.js'), content);
 }
 
-describe('hot module reload', () => {
+// This test has been skipped, because it's proven to be flaky
+// An attempt to fix this did not lead to any satisfying result
+// More info can be found here: https://github.com/xing/hops/pull/760
+describe.skip('hot module reload', () => {
   let url;
 
   beforeAll(async () => {


robin-drexler:
ok, 30 second timeout did not fix it either. I guess we need to investigate that further. :/

robin-drexler:
After adding the errors on console warnings, this is what we got now:

```
 error in browser console: Warning: Prop `%s` did not match. Server: %s Client: %s id "two" "three"
```

This is lead me to my current approach. webpack devserver blocks requests for the `develop` build, but for node, it's the untool's [`render` middleware](https://github.com/untool/untool/blob/master/packages/webpack/lib/middlewares/render.js), which iirc does not do such things.

robin-drexler:
Also broke when only executing the one test, somehow making load as the cause more unlikely. https://travis-ci.org/xing/hops/jobs/474789846

ZauberNerd:
> This is lead me to my current approach. webpack devserver blocks requests for the `develop` build, but for node, it's the untool's [`render` middleware](https://github.com/untool/untool/blob/master/packages/webpack/lib/middlewares/render.js), which iirc does not do such things. Although I'm not 100% certain, because there's eprom and reset and stuff happening.

I don't understand why the node build is relevant here? We're only testing client-side HMR, no?
Maybe we could add something as a trigger for when we want to assert the client-side DOM state, for example we could hook into the browser's console and wait for the HMR success message or we could use a DOM mutation observer to trigger our assertion?

robin-drexler:
@ZauberNerd we are also reloading and getting an error (or warning to be precise) that server's content is `#two`, but the client's is `#three`. This seems odd to me.

robin-drexler:
After it failed once when only the one test was run, it never happened again. At least not after more than 10 retries. Trying to run all in band now.

PR #775, opened 2019-01-11
chore(hops): prefix stdout consistently with name

herschel666:
Before:
before

After:
after

### DIFF
--- a/packages/hops/lib/webpack-logger-plugin.js
+++ b/packages/hops/lib/webpack-logger-plugin.js
@@ -3,17 +3,18 @@
 const prettyMS = require('pretty-ms');
 
 class HopsWebpackLoggerPlugin {
-  constructor() {
+  constructor(options) {
     this.lastHashes = {};
+    this.name = options.name;
   }
 
   apply(compiler) {
     compiler.hooks.done.tap(this.constructor.name, stats => {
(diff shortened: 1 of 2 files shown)


robin-drexler:
Not sure I like that. If the package name something is `@myscope/my-package-name`, loads of screen real estate is wasted imo. How about making it consistent by removing it entirely? Is there even any value in having it at all?

herschel666:
I would argue, that Terminal screen estate isn't a scarce resource. And of course devs want to have feedback on what Hops is about to do.

There's one thing though with long project names: the readability might suffer. But I actually don't expect this to be the case.

robin-drexler:
It's also tedious to parse the cli output, because not all output has the prefix. I don't think we can ensure every log line will have it, can we?

jhiode:
As the other log line already includes the name I think it is a good idea to do the same for all (or the other log lines that you see all the time).

robin-drexler:
@jhiode  Can you give some reasoning why you think having the prefix is useful at all?

> As the other log line already includes the name
Isn't a value in itself imo, as consistency could also be achieved by removing the prefix there as well.

ZauberNerd:
Let's close this for now until we have a more comprehensive solution in untool and re-use that then.

PR #789, opened 2019-02-05
Simplify `fetchData`

dmbch:
This PR follows up on the discussion in https://github.com/untool/untool/pull/261.

## Current state

We currently rely on the `fetchedData` property of the object passed to the `getTemplateData` hook to pass data between mixinstance methods. This appears to be rather indirect and brittle.

## Changes introduced here

In this PR, I proposed removing that indirection and instead rely on instance props (or module scoped variables).

## Outlook

At some point down the road (`untool` v2) we might want to look into getting rid of the `data` argument of `fetchData` altogether since it does not appear all that useful any longer...

## Checklist

- [ ] ~Necessary unit tests are added in order to ensure correct behavior~
- [ ] ~Documentation has been added~

### DIFF
--- a/packages/graphql/mixin.server.js
+++ b/packages/graphql/mixin.server.js
@@ -91,16 +91,7 @@ class GraphQLMixin extends Mixin {
   }
 
   fetchData(data = {}, element) {
-    return this.prefetchData(element).then(() => {
-      return {
-        ...data,
-        globals: {
-          ...(data.globals || {}),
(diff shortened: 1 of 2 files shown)


dmbch:
@ZauberNerd can we merge this now or is it still needed to test [canarist](https://github.com/xing/canarist)? It appears to be working beautifully by now, properly blocking untool/untool#263.

PR #798, opened 2019-02-14
feat(jest-preset): add mock function for importComponent

herschel666:
Wait for https://github.com/untool/untool/pull/281 to be merged & released.


ZauberNerd:
With the recent changes landed in https://github.com/untool/untool/pull/271 this PR will need some additional work

herschel666:
Not sure whether we actually want to have the `prop-types`-package in `jest-preset-hops`. Feels like it's a little out of place in there…?!? 🤔

herschel666:
Also the change applied to Untool (https://github.com/untool/untool/pull/281) doesn't fix the issue for Node v8, where `import()` is not defined, leading to failing Jest test runs. We either have to get rid of the `import` function call or shim it appropriately in Jest env. I'm undecided…

README.md
```diff
@@ -538,6 +538,40 @@ const Home = importComponent('./home', 'Home');
...
+
+  hops.importComponent = () =>
+    function ImportComponent({ render }) {
+      return <>{render({ loading: true })}</>;
+    };
+
+  return hops;
+});
+
+it('should display the loading state', () => {
```
toomuchdesign: Is the test assertion (`it()`) necessary for the scope of this example?

README.md
```diff
@@ -538,6 +538,40 @@ const Home = importComponent('./home', 'Home');
...
+**Example: check the loading state**
+
+```javascript
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import testRenderer from 'react-test-renderer';
+import App from '../';
+
+jest.mock('hops', () => {
+  const hops = jest.requireActual('hops');
```
toomuchdesign: Would it make sense providing the users with both `loading` and `error` state mocks to avoid possible future breaking changes related to how we mock/transpile this feature in tests?

packages/jest-preset/helpers/babel-plugin-import-component.js
```diff
@@ -0,0 +1,51 @@
+module.exports = ({ types: t }) => ({
```
toomuchdesign: I personally would add a brief example of what we're doing here. Just a matter of tastes :).

```suggestion
/*
 * Transforms:
 * import { importComponent } from 'hops';
 *
 * const Foo = importComponent(() => import('./foo'));
 * const Bar = importComponent('./bar');
 *
 * Into:
 * import { importComponent } from 'hops';
 *
 * const Foo = importComponent({component: () => require('./foo')});
 * const Bar = importComponent({component: () => require('./bar')});
 */
module.exports = ({ types: t }) => ({
```

README.md
herschel666: What do you mean?

packages/jest-preset/helpers/babel-plugin-import-component.js
herschel666: Hm… from my experience this team isn't too fond of code comments. Maybe somebody else can decide. 😛 Besides we can still add this later.

Btw: it transpiles to `importComponent({ component: require('./foo') })`.

README.md
toomuchdesign: I have the impression that the suggested mocking strategy exposes a few **implementation details** of the currently **underlying `loadable` component** which might change in future: _a function returning  a `<Fragment>` which renders the result of `arguments.render`_.

Would a possible refactor of `loadable` component involve updating these custom mocks?

herschel666: I guess altering this behaviour would be a breaking change we will actually try to avoid. But providing ready-made mocks on the other hand unnecessarily extends the API surface we're responsible for. So it's actually easier for us to have the users mock the behaviour. While it's imho still a reasonable effort for the people out there.

But speaking of `Fragments`: they're actually not needed here. 😅

packages/jest-preset/helpers/babel-plugin-import-component.js
jhiode: I like the idea, especially in this case because the used code is near impossible to understand without some context.

herschel666:
@toomuchdesign I added tests for the plugin. PTAL again! 🙏

packages/jest-preset/helpers/__tests__/babel-plugin-import-component.js
```diff
@@ -0,0 +1,39 @@
...
+describe('babel-plugin-import-component', () => {
+  it.each(tests)(
+    'should transpile the code correctly (fixture %s)',
+    (_, code, expected) => {
+      const result = babel.transform(code, babelOptions).code.trim();
+
+      expect(result).toBe(expected);
+    }
+  );
+});
```
toomuchdesign: What about totally dumb implementation of the same tests? It does the same but **zero magic and fun** involved. 🦊 

```js
/* eslint-env node, jest */
const { resolve, join } = require('path');
const { readFileSync: readFile } = require('fs');
const babel = require('@babel/core');

const fixtureFolder = resolve(__dirname, 'fixture');
const tests = [
  ['code-01.txt', 'expected-01.txt'],
  ['code-02.txt', 'expected-02.txt'],
  ['code-03.txt', 'expected-03.txt'],
  ['code-04.txt', 'expected-04.txt'],
];

const babelOptions = {
  plugins: [
    require.resolve('../babel-plugin-import-component'),
    require.resolve('@babel/plugin-syntax-dynamic-import'),
  ],
  babelrc: false,
};

describe('babel-plugin-import-component', () => {
  it.each(tests)(
    'should transpile the code correctly (fixture %s)',
    (codeSource, expectedSource) => {
      const code = readFile(join(fixtureFolder, codeSource), 'utf8');
      const expected = readFile(join(fixtureFolder, expectedSource), 'utf8');
      const result = babel.transform(code, babelOptions).code;

      expect(result.trim()).toBe(expected.trim());
    }
  );
});

```

PR #800, opened 2019-02-21
docs: add proper home page urls

robin-drexler:
Add homepage field for each package so that `npm docs hops-package` will
lead to the the appropriate readme instead of the main entry.

I deliberately omitted `hops`, because I think the main readme makes sense their.
The templates are used as, well, templates so I did not think adding our repo as homepage makes sense there either.

### DIFF
--- a/packages/jest-preset/package.json
+++ b/packages/jest-preset/package.json
@@ -40,5 +40,6 @@
   },
   "devDependencies": {
     "jest": "^23.4.2"
-  }
-}
+  },
+  "homepage": "https://github.com/xing/hops/tree/master/packages/jest-preset#readme"
+} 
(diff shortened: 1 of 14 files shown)


packages/jest-preset/package.json
```diff
@@ -40,5 +40,6 @@
   },
   "devDependencies": {
     "jest": "^23.4.2"
-  }
-}
+  },
+  "homepage": "https://github.com/xing/hops/tree/master/packages/jest-preset-hops#readme"
```
jhiode:
```diff
-https://github.com/xing/hops/tree/master/packages/jest-preset-hops#readme
+https://github.com/xing/hops/tree/master/packages/jest-preset#readme
```

jhiode:
```suggestion
  "homepage": "https://github.com/xing/hops/tree/master/packages/jest-preset#readme"
```

Issue #360, opened 2018-01-31
Integrate TypeScript Typings in Hops Repo


dmbch:
## This is my intent (choose one)

- [ ] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

Some projects have written their own TypeScript typings for Hops, but keep them in their respective repositories.

## Proposed solution

Publish TypeScript typings together with Hops (as described in: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)

We could use these typing as a starting point: https://github.com/foosbyte/kicker-trainer/tree/hops/typings

DJCordhose:
Once we have moved code to ES6 we might also consider using ts-check for all the code as described here: https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files

KingHenne:
In addition to the typings in https://github.com/foosbyte/kicker-trainer/tree/hops/typings this could be a starting point for `hops-redux`:

```ts
declare module 'hops-redux' {
  import {
    Options as ReactOptions,
    ReactContext,
    TemplateData,
  } from 'hops-react';
  import {
    GenericStoreEnhancer,
    Middleware,
    Reducer,
    ReducersMapObject,
    Store,
    StoreEnhancer,
    compose,
  } from 'redux';

  export interface Options {
    middlewares?: Middleware[];
    reducers?: ReducersMapObject;
  }

  export class ReduxContext<State> {
    public options: Options & ReactOptions;
    public reducers: ReducersMapObject;
    public composeEnhancers: typeof compose;
    public constructor(options?: Options & ReactOptions);
    public registerReducer<S>(namespace: string, reducer: Reducer<S>): void;
    public getStore(): Store<State>;
    public createStore(): Store<State>;
    public createEnhancer(): StoreEnhancer<State>;
    public composeMiddlewares(): StoreEnhancer<State>;
    public applyMiddlewares(): GenericStoreEnhancer[];
    public getMiddlewares(): Middleware[];
    public enhanceElement<P1, P2>(
      reactElement: React.ReactElement<P1>
    ): Promise<React.ReactElement<P2>>;
    public getTemplateData<P>(
      templateData: Partial<TemplateData>,
      rootElement: React.ReactElement<P>
    ): Promise<TemplateData>;
  }

  export function createContext(options?: Options & ReactOptions): ReactContext;
}
```

DJCordhose:
[...]

PR #804, opened 2019-02-26
feat: add typescript typings for runtime API and options

ZauberNerd:
This commit adds TypeScript typings for runtime API in `hops` and
`hops-pwa` packages.

This pull request closes issue #360

### DIFF
--- a/packages/hops/lib/runtime.d.ts
+++ b/packages/hops/lib/runtime.d.ts
@@ -1,0 +1,84 @@
+import * as React from 'react';
+import { BrowserRouterProps } from 'react-router-dom';
+
+declare module 'hops' {
+  type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+  export class Miss extends React.Component {}
+
+  export interface StatusProps {
+    code: number;
+  }
+  export class Status extends React.Component<StatusProps> {}
+
+  export interface HeaderProps {
+    name: string;
+    value: string;
+  }
+
+  export class Header extends React.Component<HeaderProps> {}
+
+  export interface ImportComponentRenderOptions<Props> {
+    Component: React.ComponentType<Props>;
+    error: boolean;
+    loading: boolean;
+  }
+
+  export interface ImportComponentOptions<Props> {
+    loader?: (load: Promise<any>) => Promise<any>;
+    render?: (
+      options: ImportComponentRenderOptions<Props> & Props
+    ) => React.ReactNode;
+  }
+
+  export function importComponent<Props>(
+    load: () => Promise<{ default: React.ComponentType<Props> }>
+  ): React.ComponentType<ImportComponentOptions<Props> & Props>;
+
+  export function importComponent<Props, Exports>(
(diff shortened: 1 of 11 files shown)


packages/hops/lib/runtime.d.ts
```diff
@@ -0,0 +1,83 @@
+import { BrowserRouterProps } from 'react-router-dom';
+
+declare module 'hops' {
+  type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+  export class Miss extends React.Component<any, any> {}
+
+  export interface StatusProps {
+    code?: number;
```
herschel666: `code` is a required prop & also `staticContext` seems to be missing.

packages/hops/lib/runtime.d.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  export interface Config {
+    [key: string]: any;
+  }
+
+  export const ServerDataContext: React.Context<ServerData>;
+
+  export const ConfigContext: React.Context<Config>;
+
+  export function withServerData<
```
herschel666: This looks quite complicated. Wouldn't it do the trick to augment a component of type `React.ComponentType<P>` to `React.ComponentType<P & C>`?

packages/hops/lib/runtime.d.ts
```diff
@@ -0,0 +1,83 @@
+import { BrowserRouterProps } from 'react-router-dom';
+
+declare module 'hops' {
+  type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+  export class Miss extends React.Component<any, any> {}
```
matthias-reis: Still a typescript newbie but shouldn't it be `React.Component<{}, any>` - no props allowed?

ZauberNerd: I've removed the `any`s now, so that we are using the default values, which are `{}`.

packages/hops/lib/runtime.d.ts
ZauberNerd: I changed code to be required. `staticContext` is injected and therefore not required to pass in from the outside, that's why I left it out of this definition.

packages/hops/lib/runtime.d.ts
ZauberNerd: I don't think so, because you are not supposed to pass in `serverData` as a prop yourself, but it is a HoC, therefore it should be available in the wrapped component, but should not be marked as a required prop for you to pass into the component.

packages/hops/lib/runtime.d.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  export interface ImportComponentOptions<Props> {
+    loader?: (load: Promise<any>) => Promise<any>;
+    render?: (
+      options: ImportComponentRenderOptions<Props> & Props
+    ) => React.ReactNode;
+  }
+
+  export function importComponent<Props, Exports>(
```
herschel666: The `Exports` interface is not used in that type declaration.

PR #809, opened 2019-03-01
Enable client state for apollo

robin-drexler:
Luckily with apollo client 2.5, you don't need to use apollo-link-state anymore. The client resolvers have been brought to the core package.
Upgrading it is sufficient to get that feature.
https://blog.apollographql.com/announcing-apollo-client-2-5-c12230cabbb7


robin-drexler:
I've rebased and updated this branch, because it got outdated due to other merges.

@herschel666 PTALA :)

PR #812, opened 2019-03-01
add cli flag to analyze client bundle

robin-drexler:
Adds `--analyze-client-bundle` cli flag to `build` command.
Using it will add `webpack-bundle-analyzer` allowing teams to, well, analyze their bundles.

I picked the name `analyze-client-bundle` to allow to potentially add `analyze-server-bundle` at a later point. Imo there will never be more than two kinds of bundles to analyze so I considered two bools better and easier than using an enum.

Currently no options are passed to  the `webpack-bundle-analyzer`, which causes a server to be started and the analyzer page to be opened automatically.
I'm not sure what's best here. We could also configure it to write an .html file somewhere that users can open themselves. This would allow to add more stuff to the `analyze-client-bundle` flag, but would also come with less convenience.

What do you say? Do we want such a feature at all? And in what configuration?

### DIFF
--- a/packages/hops/mixins/analyzer/mixin.core.js
+++ b/packages/hops/mixins/analyzer/mixin.core.js
@@ -1,0 +1,28 @@
+const { Mixin } = require('hops-mixin');
+
+module.exports = class AnalyzerCoreMixin extends Mixin {
+  configureCommand(definition) {
+    if (definition.command === 'build') {
+      definition.builder.analyzeClientBundle = {
+        type: 'boolean',
+        default: false,
+        describe: 'Opens webpack bundle analyzer page after client side build',
+      };
+    }
+  }
+  handleArguments(argv) {
+    this.options = { ...this.options, ...argv };
+  }
+  configureBuild(webpackConfig, loaderConfigs, target) {
+    if (
+      !['build', 'develop'].includes(target) ||
+      !this.options.analyzeClientBundle
+    ) {
+      return;
+    }
+    const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
+      .BundleAnalyzerPlugin;
+
+    webpackConfig.plugins.push(new BundleAnalyzerPlugin());
+  }
+};
--- a/packages/hops/mixins/analyzer/mixin.core.js
+++ b/packages/hops/mixins/analyzer/mixin.core.js
@@ -6,7 +6,8 @@ module.exports = class AnalyzerCoreMixin extends Mixin {
       definition.builder.analyzeClientBundle = {
         type: 'boolean',
         default: false,
-        describe: 'Opens webpack bundle analyzer page after client side build',
+        describe:
+          'Start an HTTP server displaying a bundle report visualization of client side build',
       };
     }
   }
@@ -23,6 +24,10 @@ module.exports = class AnalyzerCoreMixin extends Mixin {
     const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
       .BundleAnalyzerPlugin;
 
-    webpackConfig.plugins.push(new BundleAnalyzerPlugin());
+    webpackConfig.plugins.push(
+      new BundleAnalyzerPlugin({
+        openAnalyzer: false,
+      })
+    );
   }
 };
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ This command will execute a single build of all your assets.
 - `npm run build -- --production` builds all assets in production mode (which includes minification, etc).
 - `npm run build -- --static` export static HTML pages for all configured [locations](#default-settings) in development mode.
 - `npm run build -- --production --static` export static HTML pages in production mode (enables minification, etc).
+- `npm run build -- --analyze-client-bundle` visualize bundles' contents with [webpack bundle analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer).
 
 #### `npm run serve`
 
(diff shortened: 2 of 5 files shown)


toomuchdesign:
I was wondering whether it was more appropriate associating `--analyze-client-bundle` to the `start` command, instead of `build`. Any idea?

Beside that, staring at my bundles was always a very teaching activity in my experience and I would definitely provide this flag as a dev tool.

robin-drexler:
> If we shipped it as a part of the main hops package it means that we'd need a major hops release to extract it to a standalone package, right?

I don't think so, as it does not bring any public api.
We could introduce a new package and add this as dependency to the `hops` main package without any issues and breaking changes.


> I was wondering whether it was more appropriate associating --analyze-client-bundle to the start command, instead of build. Any idea?

`build` seemed to be perfectly suitable to me. 
Also, to get the production bundle info you'd have to use `start -p` which also causes a `serve -p` which often does not run out of the box on local machines. (missing dev proxy and env vars etc.)

toomuchdesign:
👍 From my point of view we can merge it. The feature would also probably benefit from a few lines of documentation. I will take care of it.

herschel666:
I think it's a good idea to provide easier access to the analysis of the build artifact.

But I'm not really happy with the solution. It feels like this belongs into its own package, that adds a `analyze-client-bundle`-command to the Hops CLI, provides sane defaults (I'm not a friend of automatically opening the browser from within the terminal … 🤷‍♂️) and takes (and passes on) the options of the plugin.

I would also not include this into the main Hops package right from the beginning. But rather get a feel for how people are using it.

robin-drexler:
I don't have much of an issue with putting into into its own package (although imo it's not warranted yet).

However, I'd not create a new command. In its current build you can analyze any build you want. -p -s whatever. It just works.

If we add a new command we'd have to duplicate all the build command flags.

Passing the other flags along to the plug-in only makes sense if never want to add anything else to this analyze flag/command than this very plugin.

I'm fine with not starting the server / opening the browser.

toomuchdesign:
Separate bundle or not? I'd really like to see some analysing tools integrated in Hops.
Any further idea about it @ZauberNerd and @dmbch?

packages/hops/mixins/analyzer/mixin.core.js
```diff
@@ -0,0 +1,28 @@
...
+    if (
+      !['build', 'develop'].includes(target) ||
+      !this.options.analyzeClientBundle
+    ) {
+      return;
+    }
+    const BundleAnalyzerPlugin = require('webpack-bundle-analyzer')
+      .BundleAnalyzerPlugin;
+
+    webpackConfig.plugins.push(new BundleAnalyzerPlugin());
```
ZauberNerd: I'm also not a big fan of opening the report directly in the browser, because we don't do that with our server either (as opposed to cra).
So I'd say we should set `openAnalyzer` to false.

ZauberNerd:
> I picked the name `analyze-client-bundle` to allow to potentially add `analyze-server-bundle` at a later point. Imo there will never be more than two kinds of bundles to analyze so I considered two bools better and easier than using an enum.

That's not entirely true, we already have more bundles than that, for example the service worker ;)

> Currently no options are passed to the `webpack-bundle-analyzer`, which causes a server to be started and the analyzer page to be opened automatically.
> I'm not sure what's best here. We could also configure it to write an .html file somewhere that users can open themselves. This would allow to add more stuff to the `analyze-client-bundle` flag, but would also come with less convenience.

I'd prefer it to not open a browser automatically, since we don't do that with our server too.
I'm unsure about server vs static files - with static files we could write all reports at the same time and users could just pick which one to open. When we use the server, we can only analyze the results of one compiler (node, build, service worker, etc), but it feels simpler to just point the users to a url instead of presenting them with a file path.
I mean we could also set the log level of this plugin very high and print our own message about where to find and how to open the report, for example by printing a `file://` url, we could have a url in the terminal that's actually clickable instead of just a file path, right?
 
> What do you say? Do we want such a feature at all? And in what configuration?

:+1: :heart:

ZauberNerd:
> Separate bundle or not? I'd really like to see some analysing tools integrated in Hops.
> Any further idea about it @ZauberNerd and @dmbch?

I don't mind it being in the main hops package.

packages/hops/mixins/analyzer/mixin.core.js
```diff
@@ -0,0 +1,32 @@
+const { Mixin } = require('hops-mixin');
+
+module.exports = class AnalyzerCoreMixin extends Mixin {
+  configureCommand(definition) {
+    if (definition.command === 'build') {
+      definition.builder.analyzeClientBundle = {
+        type: 'boolean',
+        default: false,
+        describe: 'Opens webpack bundle analyzer page after client side build',
```
ZauberNerd: This description is still mentioning that the analyzer will be opened automatically - that should be changed.

toomuchdesign: Good catch..!

robin-drexler:
Thanks everyone for pursuing this PR!

I'm just not really sure if pushing onto somebody else's PR without giving them a chance to do it themselves first is a common and good practice. 

I personally don't mind in this case, just wanted to leave my concerns for future contributions. 

Have a great day!

toomuchdesign:
Thank's @robin-drexler!
We all attribute great importance to intellectual property and open source best practices at Xing.

I decided to directly operate on the branch you opened only because we're both member of the same team at Xing and it was the easier way to ship this feature to our users.

Any other external contribution we'll be welcomed with full-open source honours! 😄

PR #814, opened 2019-03-04
GraphQL middleware fixes

jhiode:
## Current state

The Apollo server gets recreated for every request which is not needed. A weird side effect of that is that the request object never changes.

## Changes introduced here

As we know the path it is not required to have other phases before this handler. The Apollo server gets created only once now (initially and via every HMR).

## Checklist

### DIFF
--- a/packages/graphql/lib/mock-server-middleware.js
+++ b/packages/graphql/lib/mock-server-middleware.js
@@ -4,31 +4,31 @@ import express from 'express';
 import hopsConfig from 'hops-config';
 import schema from 'hops-graphql/schema';
 
-const app = express();
-app.use(cookieParser());
-
-const schemaPromise = Promise.resolve(
+const apolloAppPromise = Promise.resolve(
   typeof schema === 'function' ? schema() : schema
-);
+).then(resolvedSchema => {
+  const app = express();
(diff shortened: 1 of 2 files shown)


packages/graphql/lib/mock-server-middleware.js
```diff
@@ -4,31 +4,31 @@ import express from 'express';
...
-      cors: {
-        credentials: true,
-        origin: true,
-        methods: 'GET,HEAD,PUT,PATCH,POST,DELETE',
-        allowedHeaders: 'content-type',
-      },
-    });
+  const server = new ApolloServer({
+    schema: resolvedSchema,
+    context: context => ({ ...context, config: hopsConfig }),
```
herschel666: So this now exposes at least the `Response` object to the resolvers. Do we really want this?

jhiode: That was the idea but reading the documentation it seems that we only have the req object there.

> When context is defined as a function, it will be called on each request and receive an object containing a req property which represents the request.

jhiode:
Hey @herschel666, I changed the commit message of the first comment to be of kind `feat` in order to trigger a minor release because the middleware phase has changed now.

PR #816, opened 2019-03-07
chore: pin Node versions to major LTS ones

herschel666:
Hops currently only requires a minimum Node version, namely `8.10`. But this is actually not what we want, because that includes the non-LTS versions `9` and `11`.

So this PR introduces the more specific requirement of `^8 || ^10` for all packages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ install:
 
 matrix:
   include:
-    - node_js: '8.10'
-      name: 'Test on Node.js 8.10'
+    - node_js: '8'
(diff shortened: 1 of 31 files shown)


dmbch:
@herschel666 do we only support LTS versions? Why not support `current` (i.e. 11) as well?

For reference: https://nodejs.org/en/about/releases/

herschel666:
As discussed today IRL I close this PR now because it will definitely introduce a breaking change. Also testing on `8.10` instead of latest `8` in Travis makes sense, since that's the version that's used in the AWS Lambda runtime.

Issue #821, opened 2019-03-12
Provide `location` object as `redux.actionCreators.action` argument


toomuchdesign:
## This is my intent (choose one)

- [ ] I want to report a bug
- [X] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem
`redux.actionCreators.action` functions are currently [executed with a single argument](https://github.com/xing/hops/blob/b92c70934268918be084e38e8603b73af75a7574/packages/redux/action-creator-dispatcher/mixin.runtime-common.js#L72): `params`.

`params` is the result of `react-router`'s [`matchPath` utility](https://github.com/ReactTraining/react-router/blob/a4d0064a328477b899bb89c58dedb53e28f395a1/packages/react-router/modules/__tests__/matchPath-test.js).

In some cases this information is not enough.

## Proposed solution
In order to let **action creators** have more information about the **current request's url**, we might consider providing the whole **`location` object as second argument**.

toomuchdesign:
Closed by #829.

PR #829, opened 2019-03-18
Add second parameter "location" to actions in "actionCreators"

EricLambrecht:
This pull request closes issue #821 

As already discussed and proposed in #821 this could be a very useful feature.

## Current state

Currently actions in the `actionCreators` array only receive one parameter (`params`). 

## Changes introduced here

Actions will receive an additional parameter: an object containing the `location` and `match` objects from `react-router`. This enables the action creator to access to the search query, the current pathname, matched path etc.

`createLocation` utility gets now called with [`request.originalUrl`](https://expressjs.com/en/api.html#req.originalUrl) instead of the original [`request.path`](https://expressjs.com/en/api.html#req.path).

I also added a new paragraph to the documentation to explain the (currently undocumented) parameters that actions receive in `actionCreators`. The code-example has also been updated to make the use of the `params` more explicit.

## Checklist

### DIFF
--- a/packages/redux/README.md
+++ b/packages/redux/README.md
@@ -88,6 +88,8 @@ When server-side rendering/data fetching is enabled, this will dispatch matching
 
 On the client-side by default this will dispatch matching actions only on client-side navigation (can be overriden by setting `alwaysDispatchActionsOnClient` to `true`).
 
+Actions receive two parameters: `params` (see [URL Parameters](https://reacttraining.com/react-router/web/example/url-params) in the react-router docs) and `location` (the react router [`location`](https://reacttraining.com/react-router/web/api/location) object).
+
 ```javascript
 export default render(<MyApp />, {
   redux: {
@@ -100,7 +102,7 @@ export default render(<MyApp />, {
       },
       {
         path: '/users/:id',
-        action: fetchUser,
+        action: id => fetchUser(id),
       },
     ],
   },
--- a/packages/redux/README.md
+++ b/packages/redux/README.md
@@ -88,7 +88,7 @@ When server-side rendering/data fetching is enabled, this will dispatch matching
 
 On the client-side by default this will dispatch matching actions only on client-side navigation (can be overriden by setting `alwaysDispatchActionsOnClient` to `true`).
 
-Actions receive two parameters: `params` (see [URL Parameters](https://reacttraining.com/react-router/web/example/url-params) in the react-router docs) and `location` (the react router [`location`](https://reacttraining.com/react-router/web/api/location) object).
+Actions receive two parameters: `params` (see [URL Parameters](https://reacttraining.com/react-router/web/example/url-params) in the react-router docs) and an object containing `location` (the react router [`location`](https://reacttraining.com/react-router/web/api/location) object) and [`match`]((https://reacttraining.com/react-router/web/api/match).
 
 ```javascript
 export default render(<MyApp />, {
--- a/packages/redux/README.md
+++ b/packages/redux/README.md
@@ -102,7 +102,7 @@ export default render(<MyApp />, {
       },
       {
         path: '/users/:id',
-        action: id => fetchUser(id),
+        action: ({ id }) => fetchUser(id),
       },
     ],
   },
--- a/packages/redux/action-creator-dispatcher/mixin.runtime-common.js
+++ b/packages/redux/action-creator-dispatcher/mixin.runtime-common.js
@@ -69,7 +69,7 @@ class ReduxActionCreatorRuntimeMixin extends Mixin {
         if (!match) {
           return Promise.resolve();
         }
-        return store.dispatch(action(match.params));
+        return store.dispatch(action(match.params, location));
       })
     );
   }
--- a/packages/redux/action-creator-dispatcher/mixin.runtime-common.js
+++ b/packages/redux/action-creator-dispatcher/mixin.runtime-common.js
@@ -69,7 +69,7 @@ class ReduxActionCreatorRuntimeMixin extends Mixin {
         if (!match) {
           return Promise.resolve();
         }
-        return store.dispatch(action(match.params, location));
+        return store.dispatch(action(match.params, { location, match }));
       })
     );
   }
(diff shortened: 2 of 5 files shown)


packages/redux/action-creator-dispatcher/mixin.runtime-common.js
```diff
@@ -69,7 +69,7 @@ class ReduxActionCreatorRuntimeMixin extends Mixin {
         if (!match) {
           return Promise.resolve();
         }
-        return store.dispatch(action(match.params));
+        return store.dispatch(action(match.params, location));
```
toomuchdesign: In order to stay open to future API additions, I'd suggest to provide all **`actionCreators` arguments as a second args object** (keeping `match.params` as first argument to stay backward compatible).

```js
return store.dispatch(action(match.params, {
  match,
  location,
}));
```
In the next major release we might consider removing `match.params` as argument.

EricLambrecht: Alright, I refactored it.

EricLambrecht:
I added integration tests. However, for some reason the location object didn't work as expected in the current test environment. I wanted to add a test where I use the `location.search` property, but it was always empty, even when I navigated to something like "/location-test?test=foobar". 

**For example:** when doing `page.goto("/location-test?test=foobar#hash")` in a test,
the location object was as follows:
```
{ 
  pathname: "/location-test",
  search: "",
  hash: ""
}
```

Maybe this should be investigated.

packages/redux/action-creator-dispatcher/mixin.runtime-common.js
EricLambrecht: Also updated docs and PR description

EricLambrecht:
I replaced `request.path` with `request.originalUrl` to fix the location object (it was previously missing `search` and `hash`).

packages/spec/integration/redux/__tests__/develop.js
```diff
@@ -34,7 +34,20 @@ describe('redux developmet server', () => {
...
 
-    const count = await getInnerText('output');
+    const count = await getInnerText('counter');
+
+    expect(count).toBe('1');
+
+    await page.close();
+  });
+
+  it('executes route-matching action creators also if a search query or hash is present', async () => {
```
toomuchdesign: I'd personally like to wrap these `route-matching action creators` tests into a `describe` declaration, like. How do like it?

```js
describe('route-matching action creators')
  it('executes it when url contains search query or hash')
  it('executes it with fetch')
  it('provides it with expected "param" string')
  it('provides it with the expected  "match" object')
  it('provides it with the expected  "location" object')
```

packages/redux/README.md
```diff
@@ -88,6 +88,8 @@ When server-side rendering/data fetching is enabled, this will dispatch matching
 
 On the client-side by default this will dispatch matching actions only on client-side navigation (can be overriden by setting `alwaysDispatchActionsOnClient` to `true`).
 
+Actions receive two parameters: `params` (see [URL Parameters](https://reacttraining.com/react-router/web/example/url-params) in the react-router docs) and an object containing `location` (the react router [`location`](https://reacttraining.com/react-router/web/api/location) object) and [`match`]((https://reacttraining.com/react-router/web/api/match).
```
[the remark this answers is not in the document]
EricLambrecht: Personally, I like a more "human-readable" approach, but I can change that, if you want.

packages/spec/integration/redux/index.js
```diff
@@ -48,6 +73,18 @@ export default render(<ConnectedCounter />, {
         path: '/increment-fetch',
         action: incrementFetch,
       },
+      {
+        path: '/param/:param',
```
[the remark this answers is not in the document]
EricLambrecht: :D okay

packages/redux/README.md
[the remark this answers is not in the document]
EricLambrecht: They are also referred to as "actions" in the existing paragraphs BTW.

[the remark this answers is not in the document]
EricLambrecht: So maybe this should be changed in general, but in a different PR (or commit) I would suggest.

packages/spec/integration/redux/__tests__/develop.js
```diff
@@ -23,32 +23,84 @@ describe('redux developmet server', () => {
...
+      await page.close();
+    });
+
+    it('will be provided with the expected "match" object', async () => {
+      const { page, getInnerText } = await createPage();
+      await page.goto(url + '/match-test/foobar', {
+        waitUntil: 'networkidle2',
+      });
+
+      const param = await getInnerText('value');
```
toomuchdesign: Same here. Something like: `matchParams`?

packages/spec/integration/redux/__tests__/develop.js
```diff
@@ -23,32 +23,84 @@ describe('redux developmet server', () => {
...
+      await page.close();
+    });
+
+    it('will be provided with the expected "location" object', async () => {
+      const { page, getInnerText } = await createPage();
+      await page.goto(url + '/location-test?foo=bar', {
+        waitUntil: 'networkidle2',
+      });
+
+      const param = await getInnerText('value');
```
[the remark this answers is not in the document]
EricLambrecht: Yes, great idea!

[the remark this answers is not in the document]
EricLambrecht: Done

packages/spec/integration/redux/__tests__/develop.js
EricLambrecht: Done

packages/spec/integration/redux/index.js
```diff
@@ -24,15 +32,32 @@ const incrementFetch = () => dispatch => {
...
+    type: 'SET_VALUE',
+    payload: matchParams.test,
+  };
+};
+
+const Counter = ({ count, increment, val }) => (
   <React.Fragment>
     <button onClick={increment}>+</button>
-    <output>{count}</output>
+    <counter>{count}</counter>
```
herschel666: `<counter>` and `<value>`? Interesting choice of elements here.

Issue #831, opened 2019-03-19
hops-redux: not compatible with history v4.9


toomuchdesign:
## This is my intent (choose one)

- [X] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

`hops-redux` [imports a `createLocation` utility](https://github.com/xing/hops/blob/5dbcce1308953513c95fbcd2218793d8c691ee7f/packages/redux/action-creator-dispatcher/mixin.server.js#L2) from [history package](https://github.com/ReactTraining/history) which from v4.9 is no more accessible.

## Proposed solution

Since `createLocation` is not provided as exported entry point we might be no more able to consume it from `history` package.

- Find an alternative standalone library
- Bring the utility into Hops
- See if `history`'s maintainers agree on supporting public `DOMUtils` and `LocationUtils`. See how [this issue](https://github.com/ReactTraining/history/issues/682) evolves

toomuchdesign:
Fixed by #832.

PR #832, opened 2019-03-19
fix(redux): import "createLocation" from "history"

toomuchdesign:
This pull request closes issue #831.

## Current state

`hops-redux` [imports a `createLocation` utility](https://github.com/xing/hops/blob/5dbcce1308953513c95fbcd2218793d8c691ee7f/packages/redux/action-creator-dispatcher/mixin.server.js#L2) from [history package](https://github.com/ReactTraining/history) which from v4.9 is no more directly accessible as submodule.

## Changes introduced here

Import `createLocation` from `history` as named export.

## Notes
[history package](https://github.com/ReactTraining/history) doesn't provide **official documentation** about **accessing `createLocation`** utility.

`createLocation` is **currently available** as named export in both `history` **v4.7.2** and **v4.9**, but as long it's **not officially documented**, it might get potentially removed.

Should address #828 & #830.

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -686,7 +686,7 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
-"@babel/runtime@^7.3.1":
+"@babel/runtime@^7.1.2", "@babel/runtime@^7.3.1":
   version "7.3.4"
   resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.3.4.tgz#73d12ba819e365fcf7fd152aed56d6df97d21c83"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -5858,15 +5858,16 @@ hide-powered-by@1.0.0:
   integrity sha1-SoWtZYgfYoV/xwr3F0oRhNzM4ys=
 
 history@^4.7.2:
```
jhiode: I think we should increase the minimal required version in the package.json to `^4.9.0` as well.

PR #838, opened 2019-03-29
Implement preflight checks and config validation

ZauberNerd:


packages/development-proxy/preset.js
```diff
@@ -1,3 +1,28 @@
 module.exports = {
   mixins: [__dirname],
+  configSchema: {
```
dmbch: I like this schema a lot - very strict and explicit.

packages/graphql/mixin.core.js
```diff
@@ -41,11 +42,15 @@ class GraphQLMixin extends Mixin {
...
+                if (typeof this.getLogger === 'function') {
+                  this.getLogger.info('Fetched and saved GraphQL fragments');
+                }
               })
               .catch(err => {
-                console.error('Could not fetch GraphQL fragments:');
-                console.trace(err);
+                if (typeof this.getLogger === 'function') {
+                  this.getLogger.error('Could not fetch GraphQL fragments:');
+                  console.trace(err);
```
dmbch: I would prefer us to come up with a different solution here - maybe just call  `this.getLogger().error()` twice? It does have `Error` object handling built in, after all.

packages/graphql/mixin.core.js
```diff
@@ -41,11 +42,15 @@ class GraphQLMixin extends Mixin {
               headers: argv.header,
             })
               .then(() => {
-                console.log('Fetched and saved GraphQL fragments');
+                if (typeof this.getLogger === 'function') {
+                  this.getLogger.info('Fetched and saved GraphQL fragments');
```
dmbch: This needs to be `this.getLogger()`.

packages/lambda/mixin.core.js
```diff
@@ -69,6 +85,54 @@ class LambdaMixin extends Mixin {
...
+        semver.coerce(targetNodeVersion),
+        semver.coerce(MAX_NODE_VERSION)
+      )
+    ) {
+      warnings.push(
+        'AWS Lambda only supports Node.js up to version:',
+        MAX_NODE_VERSION
+      );
+      warnings.push(
+        'Please specify or use a Node.js version lower than or equal to this',
```
dmbch: These lines should not be separate warning, but rather concatenated - this applies to all similar occurrences in this file...

packages/spec/integration/development-proxy/__tests__/object-config.js
```diff
@@ -24,19 +25,21 @@ describe('development proxy object config', () => {
   });
 
   it('proxies with proxy config set as object', async () => {
-    expect(await fetch(url + '/dmbch').then(r => r.text())).toBe('proxy:dmbch');
-    expect(await fetch(url + '/zaubernerd').then(r => r.text())).toBe(
+    expect(await fetch(urlJoin(url, '/dmbch')).then(r => r.text())).toBe(
```
dmbch: I did not even know I was mentioned in here <3

packages/development-proxy/mixin.core.js
```diff
@@ -26,14 +24,6 @@ class ProxyMixin extends Mixin {
     };
 
     if (typeof proxyConfig === 'string') {
-      if (!checkUrl(proxyConfig)) {
-        console.warn(
-          `The proxy target '${proxyConfig}' is invalid, disabling feature`
-        );
-
-        return;
```
jhiode: I think we still need to return here as this allows to have a setup like this:
```js
{
  proxy: {
    '/my/api': 'PROXY_URL'
  }
}
```

Otherwise you would have to either always define the PROXY_URL or to remove the config entry when you want to start without a proxy.

packages/development-proxy/mixin.core.js
```diff
@@ -60,15 +50,6 @@ class ProxyMixin extends Mixin {
         const { target, routes } =
           typeof config === 'string' ? { target: config } : config;
 
-        if (target === '') {
-          console.warn(
-            `The proxy target '${target}' is invalid, ignoring path`,
-            path
-          );
-
-          return;
```
jhiode: Same here

packages/lambda/lib/create-lambda-bundle.js
```diff
@@ -54,7 +55,9 @@ module.exports = function createLambdaBundle(
 
     var archive = archiver('zip', { zlib: { level: 9 } });
     archive.on('error', reject);
-    archive.on('warning', console.warn);
+    if (logger) {
+      archive.on('warning', logger.warning);
```
jhiode: Should we `.bind` this function?

packages/lambda/preset.js
```diff
@@ -10,7 +10,30 @@ module.exports = {
     stageName: 'prod',
     uniqueName: 'hops-lambda-<name>',
     cloudformationTemplateFile: path.resolve(__dirname, 'cloudformation.yaml'),
-    include: ['<cacheDir>/**'],
+    include: ['<serverDir>/**'],
```
jhiode: Out of curiosity: could this be a breaking change for someone?

ZauberNerd: I don't think so, it's a mistake on our side, because we removed `cacheDir` and added `serverDir` instead in untool.
And we have a workaround already in place in the `aws-config.js` file:
https://github.com/xing/hops/blob/doctor/packages/lambda/lib/aws-config.js#L23

ZauberNerd: I removed the workaround in the `aws-config.js` file.

packages/typescript/mixin.core.js
```diff
@@ -19,6 +21,20 @@ class TypescriptMixin extends Mixin {
...
+  diagnose() {
+    const tsConfigPath = join(this.config.rootDir, 'tsconfig.json');
+    const exampleTsConfigPath = require.resolve(
+      'hops-typescript/tsconfig.json'
+    );
+    if (!existsSync(tsConfigPath)) {
+      return `No "tsconfig.json" file found in your project root directory ("${
+        this.config.rootDir
+      }").\nAs a starting point you can copy our minimal example config file: "cp ${exampleTsConfigPath} ${
+        this.config.rootDir
```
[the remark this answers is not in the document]
ZauberNerd: That will be built in to untool with the next release: https://github.com/untool/untool/pull/322/files#diff-e42e82624afc3020654774ec5c42b6abR19

packages/development-proxy/mixin.core.js
```diff
@@ -26,11 +24,12 @@ class ProxyMixin extends Mixin {
...
     if (typeof proxyConfig === 'string') {
-      if (!checkUrl(proxyConfig)) {
-        console.warn(
-          `The proxy target '${proxyConfig}' is invalid, disabling feature`
-        );
-
+      if (proxyConfig === '') {
+        if (typeof this.getLogger === 'function') {
+          this.getLogger().info(
+            `The proxy target "${proxyConfig}" is invalid, disabling feature`
```
jhiode: I think we should write "is empty" now.

packages/development-proxy/mixin.core.js
```diff
@@ -61,11 +60,11 @@ class ProxyMixin extends Mixin {
...
 
         if (target === '') {
-          console.warn(
-            `The proxy target '${target}' is invalid, ignoring path`,
-            path
-          );
-
+          if (typeof this.getLogger === 'function') {
+            this.getLogger().info(
+              `The proxy target "${target}" is invalid, ignoring path "${path}"`
```
jhiode: same here

packages/development-proxy/mixin.core.js
```diff
@@ -26,11 +24,12 @@ class ProxyMixin extends Mixin {
     };
 
     if (typeof proxyConfig === 'string') {
-      if (!checkUrl(proxyConfig)) {
-        console.warn(
-          `The proxy target '${proxyConfig}' is invalid, disabling feature`
-        );
-
+      if (proxyConfig === '') {
```
jhiode: If a project has no proxy config at all, which value would we get here then?

packages/development-proxy/preset.js
```diff
@@ -1,3 +1,28 @@
...
+                {
+                  type: 'object',
+                  properties: { target: { type: 'string', format: 'uri' } },
+                },
+              ],
+            },
+          },
+          additionalProperties: false,
+        },
+      ],
```
jhiode: Does this make the proxy configuration required? It is optional at the moment.

ZauberNerd: No, all properties are optional unless specified otherwise (see [required properties](https://json-schema.org/understanding-json-schema/reference/object.html#required))

packages/development-proxy/mixin.core.js
ZauberNerd: I don't think that we will get that far, because we have an [early exit](https://github.com/xing/hops/pull/838/files/f9d63985c79fdd7408ed7e1dcb5a6202940793e1#diff-24db1b4dc461fdb7ce25f11f261d6020R14) if the property `proxyConfig` is falsy.

packages/development-proxy/mixin.core.js
ZauberNerd: done

packages/development-proxy/mixin.core.js
ZauberNerd: done

packages/graphql/mixin.core.js
```diff
@@ -109,6 +114,17 @@ class GraphQLMixin extends Mixin {
       );
     }
   }
+
+  diagnose() {
+    if (!existsSync(this.config.fragmentsFile)) {
+      return [
```
dmbch: We might want to use [`submitDiagnose`](https://github.com/untool/untool/blob/master/packages/info/lib/diagnoses/generic.js#L11) for use cases such as this.

packages/spec/integration/react/__tests__/develop.js
```diff
@@ -1,4 +1,5 @@
 const fetch = require('cross-fetch');
+const urlJoin = require('url-join');
 
 describe('react developmet server', () => {
```
dmbch: `developmet` => `development`

packages/graphql/mixin.core.js
ZauberNerd: Why? This should be displayed as a warning.
Or do you mean we should use that hook to be able to print a warning and a hint?
As in:

```javascript
  diagnose({ submitDiagnosis }) {
    if (!existsSync(this.config.fragmentsFile)) {
      const logger = this.getLogger();
      submitDiagnosis(
        logger.warning.bind(logger),
        `Could not find a graphql introspection query result at "${
          this.config.fragmentsFile
        }".`
      );
      submitDiagnosis(
        logger.hint.bind(logger),
        'You might need to execute "hops graphql introspect"'
      );
    }
  }
```

Issue #802, opened 2019-02-26
How do I pass data from express to react using hops?


saviomuc:
I want to use express to evaluate the user agent string (i.e. to check if a request comes from a mobile device) and pass this information to the view.

The setup looks like this:

```
mixin.core.js:

 configureServer(app, middlewares, mode) {
    app.use((req, res, next) => {
      // my code
      next();
    });
  }
```

However I struggle to understand how to pass the information I got from parsing the request into the view. I'd prefer a solution which passes props.

The documentation mentions `withServerData` but no info on how to use it. I could also mutate `this.config` and use `withConfig` hoc but that feels dirty. 
Do you have some recommendation or hint on how to achieve that?

ZauberNerd:
I would propose to store the information you need into express' `res.locals` in your core mixin and then use them with serverData in a browser/server mixin:

`mixin.core.js`
```
configureServer(app, middlewares, mode) {
  middlewares.initial.push((req, res, next) => {
    res.locals.isMobile = somethingWithRequest(req);
  });
}
```

`mixin.server.js`
```
enhanceServerData(data, req, res) {
  return { ...data, isMobile: res.locals.isMobile };
}
```

`my-component.js`
```
const MyComponent extends React.Component {
  render() {
    return <h1>{this.props.serverData.isMobile}</h1>
  }
}
export default withServerData(MyComponent);
```

saviomuc:
Nice! Love it. Do you mind if I create a PR which explains this workflow? Not sure if this is too specific but I feel other users might be interested as well. Otherwise we can close the issue.

ZauberNerd:
@saviomuc yes, that would be great.
I assume you will be adding it to the advanced examples section? https://github.com/xing/hops#configuring-expressjs-through-a-core-mixin

We also have [this bit](https://github.com/xing/hops#withserverdatacomponent) in the main readme about the `withServerData` HoC, but I suppose it should have a link pointing to: https://github.com/xing/hops/tree/master/packages/react#withserverdatacomponent-higherordercomponent and possibly to: https://github.com/xing/hops/tree/master/packages/react#enhanceserverdataserverdata-req-res-serverdata-pipe-server

saviomuc:
@ZauberNerd 

[...]

PR #847, opened 2019-04-10
docs: Add example for passing data from server to component

saviomuc:
This pull request closes issue #802 

### DIFF
--- a/README.md
+++ b/README.md
@@ -1145,7 +1145,7 @@ enhanceServerData(data, req, res) {
 **`my-component.js`**
 
 ```javascript
-const MyComponent = ({serverData}) => <div>{serverData.isMobile}</div>
+const MyComponent = ({serverData}) => <div>{serverData.someKey}</div>
 
 export default withServerData(MyComponent);
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1117,6 +1117,39 @@ class MyExpressMixin extends Mixin {
...
+```javascript
+enhanceServerData(data, req, res) {
+  return { ...data, someKey: res.locals.someKey };
+}
+```
+
+**`my-component.js`**
+
+```javascript
+const MyComponent = ({serverData}) => <div>{serverData.isMobile}</div>
```
ZauberNerd: Shouldn't this be `serverData.someKey` as seen in the example above?

saviomuc: Valid! I changed it accordingly.

jhiode:
Hey, thanks for the PR!

In my opinion we should also document the drawbacks of injected data. I have at least two in mind:
1. the response might not be cacheable anymore because the data is request-specific.
2. the injected data can get stale easily because a SPA does not get reloaded as often as a classical page.

saviomuc:
Hi @jhiode I am not sure if that's the correct scope for this.

The purpose of this PR is just to clarify the workflow to pass data from server. 
While the issues you list might be related imho they also introduce noise which is not directly related to the documentation problem I like to solve. 

I'd rather change this PR to a more simple example instead of introducing a discussion of potential problems.

PR #848, opened 2019-04-10
[RFC]: refactor: use "untool" dist package instead of discrete packages

ZauberNerd:
With this commit we replace all occurrences of @untool/* packages with
the special `untool` dist package to ensure Hops has only a single
dependency to untool.

### DIFF
--- a/packages/jest-preset/mocks/hops.js
+++ b/packages/jest-preset/mocks/hops.js
@@ -1,5 +1,7 @@
-/* eslint react/prop-types: 0 */
 const hops = require('hops/lib/runtime');
+const {
+  internal: { runtime },
+} = require('untool');
 const importComponent = require('../helpers/import-component');
 
-module.exports = Object.assign(hops, { importComponent });
+module.exports = Object.assign(hops, runtime, { importComponent });
--- a/packages/pwa/dom.js
+++ b/packages/pwa/dom.js
@@ -1,8 +1,6 @@
 /* eslint-env browser */
 
-const {
-  internal: { getConfig },
-} = require('@untool/core');
+const hopsConfig = require('hops-config');
 
 const isLocalHost = host => {
   return (
@@ -27,7 +25,7 @@ module.exports = function installServiceWorker() {
       isLocalHost(window.location.hostname)
     ) {
       window.addEventListener('load', () => {
-        resolve(navigator.serviceWorker.register(getConfig().workerPath));
+        resolve(navigator.serviceWorker.register(hopsConfig.workerPath));
       });
     }
   });
--- a/packages/pwa/lib/loader-shim.js
+++ b/packages/pwa/lib/loader-shim.js
@@ -1,1 +1,3 @@
-export { getConfig } from '@untool/core/lib/config';
+import hopsConfig from 'hops-config';
+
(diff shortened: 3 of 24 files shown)


ZauberNerd:
> You've probably already discussed about it but.. I'd expect the `@untool` wrapper package to be released under `@untool` namespace (maybe `@untool/untool`). Is there a gain for having now namespaced and non-namespaced `@untool` packages?

It is a common pattern, to have some library (eg. `untool`, `lerna`, etc) be released under their canonical name, but have extra packages namespaced with `@untool/*` or `@lerna/*`.

packages/jest-preset/mocks/hops.js
```diff
@@ -1,5 +1,5 @@
 /* eslint react/prop-types: 0 */
-const hops = require('hops/lib/runtime');
+const hops = require('untool').internal.runtime;
```
dmbch: Is this not somewhat breaking? After all, `hops/lib/runtime` exports quite a bit more than `untool`'s runtime...

packages/pwa/dom.js
```diff
@@ -2,7 +2,7 @@
 
 const {
   internal: { getConfig },
-} = require('@untool/core');
+} = require('untool');
```
dmbch: Would it make sense to use `hops-config` in this case?

packages/pwa/lib/loader-shim.js
```diff
@@ -1 +1,3 @@
-export { getConfig } from '@untool/core/lib/config';
+import { internal } from 'untool';
+
+export const getConfig = internal.getConfig;
```
dmbch: Would it make sense to use `hops-config` in this case?

packages/react/mixin.core.js
```diff
@@ -13,13 +16,13 @@ class ReactCoreMixin extends Mixin {
...
     const untoolPluginIndex = jsLoaderConfig.options.plugins.findIndex(
-      isBabelPlugin('@untool/react/lib/babel')
+      isBabelPlugin(babelPlugin.path)
     );
 
     const untoolPlugin = jsLoaderConfig.options.plugins[untoolPluginIndex];
 
     jsLoaderConfig.options.plugins[untoolPluginIndex] = [
-      require.resolve('@untool/react/lib/babel'),
+      require.resolve(babelPlugin.path),
```
dmbch: The `require.resolve` call appears to be redundant.

packages/jest-preset/mocks/hops.js
ZauberNerd: Yes, you're right. I've reverted this change.

ZauberNerd:
@dmbch @toomuchdesign please take another look

ZauberNerd:
We decided to drop this PR, because otherwise we would have to forward all peer dependencies of `untool` (`react`, `react-dom`, `react-helmet` and `react-router-dom`) to all "top level" packages of Hops (basically all Hops packages would need to forward all other peer dependencies) and this would grow worse for our internal packages as well.
Also, we do have the doctor in place to warn about duplicate packages, so we feel a bit more confident, that we will catch misaligned dependencies anyways.

PR #851, opened 2019-04-12
fix(development-proxy): allow empty proxy targets in config

jhiode:
## Current state

We allow the proxy target to be empty in order to support a case where the proxy can get optionally enabled by providing an environment variable.

Example:
```js
  proxy: {
    '/my/api': '[PROXY_URL]',
  },
```

But that produces the following warning:
```
crate:warning Invalid configuration value(s) detected:
- config.proxy should be string
- config.proxy['/my/api'] should match format "uri"
- config.proxy['/my/api'] should be object
- config.proxy['/my/api'] should match exactly one schema in oneOf
- config.proxy should match exactly one schema in oneOf
```

## Changes introduced here

An empty string is also valid now.

## Checklist

### DIFF
--- a/packages/development-proxy/preset.js
+++ b/packages/development-proxy/preset.js
@@ -12,10 +12,18 @@ module.exports = {
           patternProperties: {
             '^/': {
               oneOf: [
+                { type: 'string', maxLength: 0 },
                 { type: 'string', format: 'uri' },
                 {
                   type: 'object',
-                  properties: { target: { type: 'string', format: 'uri' } },
+                  properties: {
+                    target: {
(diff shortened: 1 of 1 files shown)


packages/development-proxy/preset.js
```diff
@@ -12,10 +12,18 @@ module.exports = {
           patternProperties: {
             '^/': {
               oneOf: [
+                { type: 'string', pattern: '^$' },
```
dmbch: Have you considered using `maxLength: 0` as opposed to the RegExp? Validation error output might be a bit more self-explanatory.

PR #870, opened 2019-05-07
[WIP]: Resolve webpack externals relative to their context

ZauberNerd:
Sometimes externals are inside nested `node_modules` folders, for example,
if `apollo-server-express` is not in the top-level
`node_modules` folder, but instead in
`node_modules/hops-graphql/apollo-server-express`.

Before this change we would just declare them as externals and then the 
node process would try to `require('apollo-server-express')` which does not
work, because `apollo-server-express` is deeper nested. Therefore the
`require` should look like this:
`require('hops-graphql/node_modules/apollo-server-express')` in order to
actually resolve to the package.

This commit changes the externals definitions from simple strings to a 
function which does exactly that.

### DIFF
--- a/packages/graphql/mixin.core.js
+++ b/packages/graphql/mixin.core.js
@@ -1,6 +1,8 @@
 const { existsSync } = require('fs');
+const { relative, join } = require('path');
 const { Mixin } = require('hops-mixin');
 const strip = require('strip-indent');
+const resolveFrom = require('resolve-from');
 const {
   internal: { createWebpackMiddleware, StatsFilePlugin },
 } = require('@untool/webpack');
@@ -12,6 +14,31 @@ function exists(path) {
   } catch (e) {
     return false;
   }
+}
+
+function contextRelativeExternals({ rootDir }, ...externals) {
+  const countSubstrings = (input, search) => {
+    return (input.match(new RegExp(search, 'g')) || []).length;
+  };
+
+  return function(context, request, callback) {
+    if (externals.includes(request)) {
+      const absolutePath = resolveFrom(context, request);
+
+      if (countSubstrings(absolutePath, 'node_modules') > 1) {
+        const relativePath = relative(
+          join(rootDir, 'node_modules'),
+          absolutePath
+        );
+
+        return callback(null, `commonjs ${relativePath}`);
+      }
+
+      return callback(null, `commonjs ${request}`);
+    }
+
+    callback();
+  };
 }
 
 class GraphQLMixin extends Mixin {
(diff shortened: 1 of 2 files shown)


packages/graphql/mixin.core.js
```diff
@@ -14,6 +15,25 @@ function exists(path) {
...
 
+function contextRelativeExternals(externals) {
+  return function(context, request, callback) {
+    if (externals.includes(request)) {
+      const absolutePath = require.resolve(request, {
+        paths: require.main.paths.concat(context),
+      });
+      if ((absolutePath.match(/node_modules/g) || []).length > 1) {
+        const relativePath = relative(
+          this.config.rootDir,
```
herschel666: `this.config` was `undefined` in my local test. Maybe it makes sense to pass that (or the `rootDir`) in as the first argument.

packages/graphql/mixin.core.js
```diff
@@ -14,6 +15,25 @@ function exists(path) {
...
+          absolutePath
+        ).replace(/^node_modules\//, '');
```
herschel666: Rather than messing with a string here, it might be more robust to make the path relative against `path.join(rootDir, 'node_modules)`.

packages/graphql/mixin.core.js
```diff
@@ -14,6 +15,25 @@ function exists(path) {
   }
 }
 
+function contextRelativeExternals(externals) {
```
herschel666: I'd personally go for `... externals` here to not have to wrap those module names into an unnecessary array.

packages/graphql/mixin.core.js
```diff
@@ -14,6 +16,23 @@ function exists(path) {
   }
 }
 
+function contextRelativeExternals({ rootDir }, ...externals) {
+  return function(context, request, callback) {
+    if (externals.includes(request)) {
+      const absolutePath = resolveFrom(context, request);
+      if ((absolutePath.match(/node_modules/g) || []).length > 1) {
```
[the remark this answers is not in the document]
ZauberNerd: No, we want to count it, because we only want to go into this branch if there are 2 or more occurrences of `node_modules` in the string. Because if there is only one, than it means that the package is directly resolvable from `rootDir`.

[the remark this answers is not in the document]
jhiode: Another way could be to write `/node_modules.*node_modules/.test(absolutePath)` but not sure if that is any better than the current solution.

ZauberNerd:
I don't think this is the correct approach right now.
In another commit I added a check for duplicate `graphql` packages, since that was the root cause for multiple installs of `apollo-server-express`.
We should monitor this and if needed revisit at a later stage.

PR #874, opened 2019-05-10
refactor(graphql): avoid duplicate tree rendering

herschel666:


packages/graphql/mixin.server.js
```diff
@@ -19,6 +23,24 @@ const fetch = require('cross-fetch');
 
 let introspectionResult = undefined;
 
+const renderMarkupAndHelmetFragments = async (element, canPrefetchOnServer) => {
+  if (canPrefetchOnServer) {
+    let helmetFragments = {};
+    const reactMarkup = await getMarkupFromTree({
+      tree: element,
+      renderFunction: tree => {
+        helmetFragments = renderStatic();
```
ZauberNerd: I think you need to change the order of these two function calls. First `renderToString` and then `renderStatic`.
https://github.com/nfl/react-helmet#server-usage

packages/graphql/mixin.server.js
```diff
@@ -19,6 +23,24 @@ const fetch = require('cross-fetch');
 
 let introspectionResult = undefined;
 
+const renderMarkupAndHelmetFragments = async (element, canPrefetchOnServer) => {
```
ZauberNerd: Shouldn't the second argument rather be `shouldPrefetchOnServer` or just `prefetchOnServer`? The question whether we *can* prefetch has already been answered through the call to the mixin hook.

packages/graphql/mixin.server.js
herschel666: I'm using the untool's implementation now, so this is fixed.

packages/graphql/mixin.server.js
herschel666: Fixed with the latest force-push.

packages/graphql/mixin.server.js
```diff
@@ -82,14 +83,18 @@ class GraphQLMixin extends Mixin {
...
-    const prefetchOnServer = this.canPrefetchOnServer().every(value => value);
-
-    return prefetchOnServer ? getDataFromTree(element) : Promise.resolve();
+  async renderToFragments(element) {
+    if (this.canPrefetchOnServer().every(value => value)) {
+      let fragments = {};
+      await getMarkupFromTree({
+        tree: element,
+        renderFunction: tree => {
+          fragments = renderToFragments(tree);
```
ZauberNerd: Shouldn't the `renderFunction` return `fragments.reactMarkup`?

packages/graphql/mixin.server.js
```diff
@@ -82,14 +83,20 @@ class GraphQLMixin extends Mixin {
...
-    const prefetchOnServer = this.canPrefetchOnServer().every(value => value);
-
-    return prefetchOnServer ? getDataFromTree(element) : Promise.resolve();
+  async renderToFragments(element) {
+    if (this.canPrefetchOnServer().every(value => value)) {
+      let fragments = {};
+      const reactMarkup = await getMarkupFromTree({
+        tree: element,
+        renderFunction: tree => {
+          const { reactMarkup, ...renderedFragments } = renderToFragments(tree);
```
jhiode: Hm, IMO we don't need all that splatting and merging:
```js
  async renderToFragments(element) {
    if (this.canPrefetchOnServer().every(value => value)) {
      let fragments = {};
      await getMarkupFromTree({
        tree: element,
        renderFunction: tree => {
          fragments = renderToFragments(tree);
          return fragments.reactMarkup;
        },
      });
      return fragments;
    }
    return renderToFragments(element);
  }
```

And what do you think about returning early instead of nesting the happy path?
```js
  myFunction() {
    if (/* check for special case */) {
      // handling special case
      return;
    }

    // start of happy path
  }
```

packages/graphql/package.json
```diff
@@ -37,13 +38,15 @@
     "graphql-tag": "^2.10.0",
     "react": "^16.4.0",
     "react-apollo": "^2.2.3",
-    "react-dom": "^16.4.0"
+    "react-dom": "^16.4.0",
+    "react-helmet": "^5.2.0"
```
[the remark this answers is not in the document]
ZauberNerd: Yes, and probably `react-router-dom` too, because these are peer dependencies of `@untool/react` and all peer dependencies that are not being satisfied by this package need to be forwarded.
https://github.com/untool/untool/blob/master/packages/react/package.json#L33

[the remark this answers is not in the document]
ZauberNerd: This isn't resolved yet - `react-router-dom` is still missing in peer and dev dependencies ;)

PR #879, opened 2019-05-14
feat(graphql): pass on GraphQL network errors transparently

herschel666:
From commit [`c6ca491`](https://github.com/xing/hops/commit/c6ca491486b28d7434c8cf1ae9200de884860266):

> To prevent every ApolloError to lead to an error 500 during server-side
rendering, a dedicated `fetchError` object is thrown in case of an
error, that takes the failed response into account.  
Thereby we're able to transparently proxy the reason for the failure,
e.g. a "429 Too Many Requests" error, and also to pass on special
headers, like the `Retry-After` header of the 429 error.

Besides: I still have to detect the correct server address for the HTTP link…


packages/spec/integration/graphql/mixin.runtime.js
```diff
@@ -0,0 +1,23 @@
...
+
+  return await fetch(newUri, options);
+};
+
+module.exports = class GraphQlMixin extends Mixin {
+  getApolloLink() {
+    return new HttpLink({
+      uri: this.config.graphqlUri,
+      // TODO: get correct server address
+      fetch: customFetch(() => this.serverAddress || 'http://localhost:8080'),
```
ZauberNerd: `this.getServerAddress()` should give you that (https://github.com/untool/untool/blob/master/packages/express/mixins/mixin.core.js#L38)

herschel666: Leads to `Error: this.getServerAddress is not a function` for me. And I don't really have an idea how to get this information asynchronously from within the core-mixin… :-|

dmbch: How about `res.locals.serverAddress = this.getServerAddress()`?

packages/spec/integration/graphql/mixin.runtime.js
```diff
@@ -2,22 +2,25 @@ const { Mixin } = require('hops-mixin');
...
+  bootstrap(_, res) {
+    this.root = res ? res.locals.url : '';
+  }
+
   getApolloLink() {
     return new HttpLink({
       uri: this.config.graphqlUri,
-      // TODO: get correct server address
-      fetch: customFetch(() => this.serverAddress || 'http://localhost:8080'),
+      fetch: customFetch(() => this.root),
```
jhiode: Why can't we just pass in the value of `this.root`?

herschel666:
I added the handling of a response with a status code of `200`, that ends up as a network error. The assumption is, that the response's content-type is wrong, so I'll opted into converting this into a `406` error.

If the content type was fine, we wouldn't end up in this code path, but rather got a `WriteError` which leads to throwing the plain error object that has been passed into the catch-block.

packages/graphql/lib/get-graphql-error.js
```diff
@@ -0,0 +1,36 @@
...
+};
+
+module.exports = error => {
+  if (!error.networkError || !error.networkError.response) {
+    return error;
+  }
+
+  const { stack, response } = error.networkError;
+  const { status: statusCode, statusText, headers: responseHeaders } = response;
+  const status = statusCode === 429 ? 429 : 500;
```
jhiode: I think we need the same logic for redux so in my opinion we should just throw a new error with all useful information that we have here.

packages/graphql/lib/get-graphql-error.js
```diff
@@ -0,0 +1,36 @@
...
+  const headers = getHeaders(status, responseHeaders);
+  const message = statusCode === 200 ? 'Not Acceptable' : statusText;
+  const fetchError = Object.assign(new Error(message), {
+    name: 'HopsGraphQlError',
```
jhiode: `HopsFetchDataError`?

packages/graphql/lib/get-graphql-error.js
```diff
@@ -0,0 +1,36 @@
...
+    original: error.networkError,
+    headers,
+    status,
+  });
+  const fetchErrorStack = fetchError.stack
```
jhiode: I think it is totally ok to keep the current stack trace. And the stack trace of `networkError` isn't that helpful either because that only shows where the fetch function got called...

packages/graphql/lib/get-graphql-error.js
herschel666: Sorry, I don't understand that. Isn't this what is happening here?

packages/graphql/lib/get-graphql-error.js
herschel666: Depends on how we want to handle it for `development` envs later on. Since ppl might be using Redux and GraphQL in parallel, it could be helpful to know the source exactly rather than to have a generic fetch error.

packages/graphql/lib/get-graphql-error.js
jhiode: We have opinionated logic here about the status code and error message. I think it would be beneficial to move that logic somewhere else in order to have the same feature for redux actions.

As we can't override this anymore I would either create a mixin hook or just don't handle the status codes here at all. Responding with a status code is responsibility of an express middleware IMO - that we want take the status code of this request into account is an implementation detail of that middleware.

herschel666:
@jhiode @ZauberNerd I made the error handling more generic now — no special handling of `429` status code & no giving special names to error objects anymore. Additionally I hooked the handling of the error, so it's rewritable and people can e.g. implement a special handling of `429` errors.

What do you think?

packages/spec/integration/graphql/__tests__/develop.js
```diff
@@ -8,9 +17,61 @@ describe('graphql development client', () => {
...
     const { page } = await createPage();
     await page.goto(url, { waitUntil: 'networkidle2' });
-
     await page.waitForSelector('#commits', { timeout: 10000 });
-
     await page.close();
   });
+
+  describe('/html', () => {
+    it('should render a 406 error page for a status 200 response', async () => {
```
jhiode:
```diff
-should render a 406
+should render a 500
```

packages/graphql/mixin.server.js
```diff
@@ -20,6 +20,18 @@ const fetch = require('cross-fetch');
 
 let introspectionResult = undefined;
 
+const getApolloError = error => {
```
jhiode: I think the name is a little bit misleading as we don't get (return) an Apollo error here. We convert or wrap it...

packages/graphql/mixin.server.js
```diff
@@ -20,6 +20,18 @@ const fetch = require('cross-fetch');
...
+const wrapNetworkError = error => {
+  if (!error.networkError || !error.networkError.response) {
+    return error;
+  }
+  const { status, statusText } = error.networkError.response;
+  const message = status === 200 ? 'Not Acceptable' : statusText;
+  const fetchError = Object.assign(new Error(message), {
+    response: error.networkError.response,
+  });
+  throw fetchError;
```
herschel666: why doesn't this return?

PR #880, opened 2019-05-15
Update README.md

FDiskas:
shorter version of init project script

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ You can use either [`npx`](https://medium.com/@maybekatz/introducing-npx-an-npm-
 To create a new Hops application run the following in a terminal:
 
 ```shell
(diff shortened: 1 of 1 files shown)


ZauberNerd:
Thank you for your pull request @FDiskas.
We chose to put `npx` in the readme, because our lowest supported Node.js version is `8.6.0` which ships bundled with `npm@5.3.0` (https://nodejs.org/en/download/releases/).

`npm init` as a project initializer has only been released with `npm@6.1.0` which is bundled with Node.js in version `10.3.0` whereas `npx` is being bundled with `npm@5.2.0` and therefore all our users will already have it installed globally.

We might reconsider this, once we drop support for Node.js version `8.6.0`, but until then I would prefer to keep it as is.

Issue #815, opened 2019-03-07
Jest broken on Node 11.11.0


dmbch:
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

Unfortunately, Jest ist broken on the lates Node version, 11.11.0, see

- https://github.com/facebook/jest/pull/8050
- https://github.com/facebook/jest/issues/8070

I only noticed this because the canarist checks in `untool` that do not use hops’ `yarn.lock` are [failing](https://travis-ci.org/untool/untool/jobs/502840277).

## Proposed solution

Upgrade to `jest@24.2.0-alpha.0` - since we are hesitant to upgrade to jest@24 in general (#780, #788, #799), we should probably consider alternatives such as updating the `engines` fields of our packages.

dmbch:
Could we use a range for the dependencies in question here, i.e.  support borh jest@23 and jest@24?

toomuchdesign:
I would also propose considering a more **conservative strategy** of declaring our **`engines` fields** by **explicitly mentioning** the **major versions** we want to support.

```js
 "engines": {
    "node": "^8.10 || 9 || 10 || 11",
    "yarn": ">=1.10.1" // Maybe also with yarn?
  },
```
It might be particularly useful as long as we're keeping pretty **strict policies about realising major versions** on our side.

Beside that `>` means that `hops` is/will be compatible with the next versions of Node or Yarn, and we don't know that, yet :).

ZauberNerd:
I asked whether it would be possible to release a fix for Jest 23, but unfortunately it is not (https://github.com/facebook/jest/pull/8050#issuecomment-471639713) so we will have to upgrade Jest on our end to v24.

The problem is, that this would be a breaking change, because we expect our users to have jest installed as a dependency in their projects, but we ship the jest-preset-hops which is currently tied to jest 23.
In our preset we have some dependencies to jest packages https://github.com/xing/hops/blob/master/packages/jest-preset/package.json#L30-L37 and since these are dependencies instead of peer dependencies, we can't specify a range on them to support both versions, therefore we have to update the preset to only support v24 and the breakage would then be on the user's side, because they will have to update their local jest dependencies to v24 too.

Unfortunately I don't see another way around this at the moment, so we'll have to discuss whether we want to delay this until the end of the year when we ship Hops 12, or whether we want to release Hops 12 now or if we want break the semver rules and publish it in Hops 11 anyways..

[...]

PR #881, opened 2019-05-15
fix(template): Fix jest bugs in react template

FDiskas:

## Current state

```npm init hops-app my-hops-app```
```cd my-hops-app```
```npm test```

```ssh
Requires Babel "^7.0.0-0", but was loaded with "6.26.3". If you are sure you have a compatible version of @babel/core, it is likely that something in your build process is loading th
e wrong version. Inspect the stack trace of this error to look for the first entry that doesn't mention "@babel/core" or "babel-core" to see what is calling Babel.
```

## Changes introduced here

How to fix:
`npx babel-upgrade --write --install`
`npm audit fix --force`

> **NOTE** And commit `pakage-lock.json` or `yarn.lock` to git for keeping versions locked and not introduce unexpected errors for end users

## Checklist

### DIFF
--- a/packages/template-react/package.json
+++ b/packages/template-react/package.json
@@ -25,7 +25,9 @@
     "react-router-dom": "^5.0.0"
   },
   "devDependencies": {
-    "jest": "^23.4.2",
+    "babel-core": "^7.0.0-bridge.0",
+    "babel-jest": "^24.8.0",
+    "jest": "^24.8.0",
     "jest-preset-hops": "^11.7.1",
     "react-test-renderer": "^16.4.2"
   },


packages/template-react/package.json
```diff
@@ -25,7 +25,9 @@
     "react-router-dom": "^5.0.0"
   },
   "devDependencies": {
-    "jest": "^23.4.2",
+    "babel-core": "^7.0.0-bridge.0",
+    "babel-jest": "^24.8.0",
```
ZauberNerd: I would prefer to remove `babel-jest` here, because we have it already as a dependency in the `jest-preset-hops`:
https://github.com/xing/hops/blob/master/packages/jest-preset/package.json#L31

Also, I would like to keep jest at version 23 for now, because unfortunately we still have some discussion points open, for how we could upgrade to jest 24 without being breaking for our users (see https://github.com/xing/hops/issues/815 and https://github.com/xing/hops/pulls?q=is%3Apr+is%3Aopen+label%3Ablocked)

I will try to raise this topic to our attention and investigate if we can find a solution where we can safely upgrade to jest 24, but until then we should stay at 23 to be on the safe side.

So in a nutshell: It will be enough to just add `babel-core@7.0.0-bridge.0` to the dev dependencies ;)

ZauberNerd:
Hi @FDiskas do you want to continue on this PR or should we take over and implement this ourselves?

FDiskas:
Yes please. I'm gone for holiday

ZauberNerd:
@FDiskas I continued your PR here: https://github.com/xing/hops/pull/887

PR #882, opened 2019-05-15
feat(contributing): Added conventionalcommits to the development envi…

FDiskas:
…ronment

No more pain in preparing right commit message. Just use `yarn commit` or `git commit` with no
message and vizard will help you. More info at npmjs.com/package/commitizen

## Current state
Wrong commit messages all over the place. Thous messages are not updated in changelog

## Checklist

- [x] Documentation has been added (no needed workflow does not changed)

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
     "release:candidate:feature": "lerna publish preminor --preid=rc --npm-tag=next",
     "release:next": "lerna publish prerelease --preid=rc --npm-tag=next",
     "reset": "git clean -dfx && yarn",
-    "fmt": "prettier --write '**/*.{js,json,css}' '**/README.md'"
+    "fmt": "prettier --write '**/*.{js,json,css}' '**/README.md'",
+    "commit": "git-cz"
   },
   "canarist": {
     "repositories": [
@@ -61,7 +62,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "prepare-commit-msg": "exec < /dev/tty && git cz --hook"
     }
   },
   "lint-staged": {
@@ -147,6 +149,7 @@
   "devDependencies": {
     "babel-eslint": "^10.0.0",
     "canarist": "^1.2.1",
+    "commitizen": "^3.1.1",
     "eslint": "^5.3.0",
     "eslint-config-prettier": "^4.0.0",
     "eslint-plugin-import": "^2.12.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -158,5 +161,10 @@
     "lerna": "^3.0.3",
     "lint-staged": "^8.0.0",
     "prettier": "^1.14.0"
+  },
+  "config": {
```
ZauberNerd: I like having the config here :+1:

package.json
```diff
@@ -147,6 +149,7 @@
   "devDependencies": {
     "babel-eslint": "^10.0.0",
     "canarist": "^1.2.1",
+    "commitizen": "^3.1.1",
```
ZauberNerd: I would prefer to just install `cz-conventional-changelog` and have the config in the `package.json`, so that people who have `commitizen` installed globally may use it with our project.

package.json
```diff
@@ -61,7 +62,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "prepare-commit-msg": "exec < /dev/tty && git cz --hook"
```
ZauberNerd: I don't think this will work on windows or other environments.
Also: I'm not entirely sure that it will be a good idea to force everyone to use `git-cz`, because some times it is just easier to write your commit messages in vi/vim or whatever your preferred commit message editor is.

Maybe we could just have the `commitizen` config in this repository and then also use https://github.com/conventional-changelog/commitlint to reject any invalid messages in case the committer is not using `git-cz`?

This is similar to [how we do it in `untool`](https://github.com/untool/untool/blob/master/package.json#L134-L161)

FDiskas:
I can test this on windows. But It should work. In this pull request there is `husky` used.

So if you use simple `git commit` it will ask couple questions and the commit message will be always correct.

And why to force it? I see couple pull requests where it is not merged because of wrong message. And for later will not be needed to comment again and again about wrong commit message. This is why you are using `prettier` - for stop talking about coding style. :)

:v:

ZauberNerd:
Hi @FDiskas as explained above, I don't want to force people to use the `git-cz` editor for the following reasons:
- having a different editor than your configured one pop up might be surprising for some people
- the `git-cz` editor does not allow for multi-line commit message bodies
- the `git-cz` editor does not comfortably allow to edit parts of your message after you have written them

Therefore I would be okay with having the commitizen config in this repository, so that people who have `git-cz` installed globally will be able to use it and for all other use cases we will have commitlint which will give feedback about the commit message format.

ZauberNerd:
@FDiskas I continued this PR in https://github.com/xing/hops/pull/888

PR #883, opened 2019-05-19
[RFC] feat(lambda): add "profile" option

herschel666:
...in order to allow users to point the AWS-SDK to a different pair of
credentials, without requiring them to set lengthy environment
variables.


ZauberNerd:
Not sure, that I like this. The profile can also be set through an environment variable `AWS_PROFILE=myprofile` which would require no special handling in our code base.

herschel666:
`serverless.com`, `claudiajs.com`, `aws-amplify.github.io` and `arc.codes` all provide a configuration option to set the profile. Ppl that have worked with those tools and don't follow the anti-pattern of using the default profile for every project, will likely look for such an option.So it's imho a valuable addition.

ZauberNerd:
I don't know why other frameworks would do it that way. Using env variables is much better for transporting projects between different developers / environments:
When specifying the profile in the source code (aka. package.json), every developer and every other CI system needs to know these profile names and have them configured in exactly the same way.
Whereas when using the `AWS_PROFILE` environment variable, it doesn't matter what names some other developer or system has for that specific profile and therefore it simplifies sharing between different people who are able to deploy the lambda function.

PR #896, opened 2019-06-11
Improve renovate config

ZauberNerd:
Pimp our renovate config with more specific rules:

- add "untool" as package to the untool monorepo group filter
- **peerDependencies** should be "widened (ie: `"react": "15 || 16"` instead of upgrading to `16`)
- **devDependencies** should be bumped (ie: an update from `prettier@^1.0.0` to `prettier@1.1.0` which would normally fall under our version range, should be updated to the new range: `"prettier": "^1.1.0"`)
- **devDependencies in the root `package.json`** should be auto merged if they are a minor or patch upgrade

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -6,7 +6,7 @@
     "enabled": true
   },
   "semanticCommitType": "chore",
-  "semanticCommitScope": "dependencies",
+  "semanticCommitScope": null,
   "packageRules": [
     {
       "packagePatterns": ["^@untool/"],
@@ -15,8 +15,7 @@
     },
     {
       "depTypeList": ["dependencies", "peerDependencies"],
-      "semanticCommitType": "fix",
-      "semanticCommitScope": "dependencies"
+      "semanticCommitType": "fix"
     },
     {
       "depTypeList": ["peerDependencies"],
--- a/renovate.json
+++ b/renovate.json
@@ -8,6 +8,7 @@
   "packageRules": [
     {
       "packagePatterns": ["^@untool/"],
+      "packageNames": ["untool"],
       "groupName": "Update untool monorepo packages"
     },
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -5,18 +5,26 @@
...
     {
+      "depTypeList": ["dependencies", "peerDependencies"],
       "semanticCommitType": "fix",
-      "semanticCommitScope": "dependencies",
-      "depTypeList": ["dependencies", "peerDependencies"]
+      "semanticCommitScope": "dependencies"
+    },
+    {
+      "depTypeList": ["peerDependencies"],
+      "rangeStrategy": "widen"
```
dmbch: 👍

renovate.json
```diff
@@ -5,18 +5,26 @@
...
       "packagePatterns": ["^@untool/"],
       "packageNames": ["untool"],
       "groupName": "Update untool monorepo packages"
     },
     {
+      "depTypeList": ["dependencies", "peerDependencies"],
       "semanticCommitType": "fix",
-      "semanticCommitScope": "dependencies",
-      "depTypeList": ["dependencies", "peerDependencies"]
+      "semanticCommitScope": "dependencies"
```
dmbch: Can we not get rid of the scope altogether?

ZauberNerd:
> Automerging is not going to work: [renovatebot/renovate#846](https://github.com/renovatebot/renovate/issues/846)

:cry: 

> Why not restrict renovate to non-office hours any more? This might prevent some conflicts, does it not?

What kind of conflicts are you talking about? I initially enabled this option, to preserve our travis-ci resources, but disabled it now, because there aren't too many PRs being opened anyways (I think renovate already restricts this to max every 2 hours).

renovate.json
ZauberNerd: Yes, we could. I'll remove it.

PR #898, opened 2019-06-12
fix: rename "browsers" option for postcss to silence deprecation warning

ZauberNerd:
browserslist


packages/postcss/mixin.core.js
```diff
@@ -38,7 +38,7 @@ const getCSSLoaderConfig = (browsers, additionalLoader) => {
           plugins: [
             postcssImportPlugin(),
             postcssPresetEnv({
-              browsers,
```
jhiode: We might still need `browsers` for older versions of postcss.

ZauberNerd:
Related: https://github.com/csstools/postcss-preset-env/issues/133

PR #905, opened 2019-06-24
Return error

jhiode:
`wrapNetworkError` should return the error instead of throwing it.

Thanks @herschel666 

## Checklist


jhiode:
As we throw the return value of `wrapNetworkError` anyway, the result/behaviour is actually the same.

PR #914, opened 2019-07-04
fix(template-graphql): update graphql query

toomuchdesign:
## Current state

The query issued by graphql demo generates an error originated by this part of the query:

```
... on GithubCommitAuthor {
   email
}
```

## Changes introduced here

Remove author email retrieval and update demo accordingly.

## Checklist

- [X] Necessary unit tests are added in order to ensure correct behavior

### DIFF
--- a/packages/template-graphql/src/home/commits.gql
+++ b/packages/template-graphql/src/home/commits.gql
@@ -14,9 +14,6 @@
             ...on GithubUser {
               login
             }
-            ...on GithubCommitAuthor {
-              name
-            }
           }
         }
       }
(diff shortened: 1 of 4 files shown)


herschel666:
It seems to me you're fighting symptoms here. The query is afaict correct, but Apollo's heuristic fragment matcher is failing for reasons I haven't fully figured out yet. If you add `fetchPolicy: "no-cache"` to the options of the query, you'll see the list of commits popping up. But as soon as Apollo's caching layer comes into play, it explodes.

Maybe we have to look for a new GraphQL API for for this example. @ZauberNerd What do you think?

ZauberNerd:
@herschel666 switching to a different API would be fine, do you know of any public GraphQL API that requires no authentication?

herschel666:
I created a ticket regarding switching to a different API & will thus close this PR. Thanks @toomuchdesign!

PR #919, opened 2019-07-15
Abstract redux-thunk middleware creation

hossam-magdy:
This pull request extracts the redux-thunk middleware creation to its own method in the redux runtime mixin. So that it is clear when the mixin is extended what to replace and what to retain.

## Changes introduced here
No big changes, just a method-extraction refactor.

## Checklist


ZauberNerd:
Thanks for your contribution!
However we prefer not to merge this change, because we want to keep the number of hooks / methods low as to not increase the API surface too much.

PR #938, opened 2019-08-30
feat(postcss): upgrade to css-loader v3

ZauberNerd:
BREAKING CHANGE: Check the breaking changes in css-loader v2 and v3

### DIFF
--- a/packages/postcss/mixin.core.js
+++ b/packages/postcss/mixin.core.js
@@ -9,9 +9,10 @@ const { join, trimSlashes } = require('pathifist');
 const getAssetPath = (...args) => trimSlashes(join(...args));
 
 const cssLoaderLocalOptions = {
-  camelCase: true,
-  modules: true,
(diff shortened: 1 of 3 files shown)


packages/postcss/mixin.core.js
```diff
@@ -9,9 +9,9 @@ const { join, trimSlashes } = require('pathifist');
 const getAssetPath = (...args) => trimSlashes(join(...args));
 
 const cssLoaderLocalOptions = {
-  camelCase: true,
-  modules: true,
-  localIdentName: '[folder]-[name]-[local]-[hash:8]',
+  modules: {
+    localIdentName: '[folder]-[name]-[local]-[hash:8]',
+  },
```
herschel666: I guess we have to add `localsConvention: 'camelCase'` here, since the default value is `asIs`, which will not convert e.g. kebab-case classnames to camel-case anymore.

PR #944, opened 2019-09-10
feat(typescript): make typescript a peer dependency

jhiode:
BREAKING CHANGE:

In order to improve flexibility we changed TypeScript to be a peer dependency now.
Please install at least v3.6.0 of the `typescript` package.

### DIFF
--- a/packages/typescript/package.json
+++ b/packages/typescript/package.json
@@ -16,8 +16,13 @@
   },
   "dependencies": {
     "hops-mixin": "^11.9.0",
-    "ts-loader": "^6.0.0",
-    "typescript": "^3.2.1"
+    "ts-loader": "^6.0.0"
+  },
+  "peerDependencies": {
+    "typescript": "^3.2.0"
+  },
+  "devDependencies": {
+    "typescript": "^3.2.0"
   },
   "homepage": "https://github.com/xing/hops/tree/master/packages/typescript#readme"
 }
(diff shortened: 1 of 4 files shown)


ZauberNerd:
Why 3.6 instead of `^3.0.0`?
Also: I think you should have a short summary after the words `BREAKING CHANGE:` - the first line will be put into the change log.

packages/spec/integration/typescript/package.json
```diff
@@ -15,6 +15,7 @@
   },
   "dependencies": {
     "hops": "*",
-    "hops-typescript": "*"
+    "hops-typescript": "*",
+    "typescript": "^3.6.0"
```
ZauberNerd: This doesn't make sense at the moment, because the spec packages are not workspaces.

jhiode:
We use 3.2 now because that allows to use a package name in the `extends` field (See https://github.com/microsoft/TypeScript/pull/27348).

packages/typescript/package.json
```diff
@@ -16,8 +16,13 @@
   },
   "dependencies": {
     "hops-mixin": "^11.9.0",
-    "ts-loader": "^6.0.0",
-    "typescript": "^3.2.1"
+    "ts-loader": "^6.0.0"
+  },
+  "peerDependencies": {
+    "typescript": "*"
```
ZauberNerd: I think it would be better to restrict the peer dependency versions here.

PR #952, opened 2019-09-27
chore(config): deprecate hops-config package

herschel666:
The commit message's body:

Instead of importing the `config`-object from a package, you should
rather use `ConfigContext` or the `withConfig()`-HOC from `hops`.

**Example using `ConfigContext`**
```jsx
import React from 'react';
import { ConfigContext } from 'hops';

const Comp = () => {
  const config = React.useContext(ConfigContext);
  return <em>{config.basePath}</em>;
};
```

**Example using `withConfig` HOC**
```jsx
import React from 'react';
import { withConfig } from 'hops';

const Comp = withConfig(({ config }) => (
  <em>{config.basePath}</em>
));
```

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -1,0 +1,5 @@
+# Deprecated APIs
+
+### DEP0001
+
+`hops-config` is deprecated. Please use `ConfigContext` or the [`withConfig()`](https://github.com/xing/hops#withconfigcomponent) HOC from `hops` instead.
--- a/packages/config/browser.js
+++ b/packages/config/browser.js
@@ -1,1 +1,7 @@
+const deprecate = require('depd')('hops-config');
+
+deprecate(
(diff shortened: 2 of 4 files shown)


DEPRECATIONS.md
```diff
@@ -0,0 +1,5 @@
+# Deprecated APIs
+
+### DEP0001
+
+`hops-config` is deprecated. Please use the [`withConfig()`](https://github.com/xing/hops#withconfigcomponent) HOC from `hops` instead.
```
jhiode: Should we instead point to the `ConfigContext` as that would also work with hooks?

ZauberNerd:
Shouldn't we also deprecate the [`index.js`](https://github.com/xing/hops/blob/master/packages/config/index.js)?

herschel666:
@ZauberNerd A deprecation message in there doesn't have any effect — why ever that's the case…?! For `browser.js` again the message shows up in the Terminal — so that's _"fine"_.

PR #966, opened 2019-10-15
test(spec): use @next version of create-hops-app

herschel666:


packages/spec/integration/create-hops-app/__tests__/index.spec.js
```diff
@@ -5,7 +5,7 @@ const { execSync } = require('child_process');
 const createHopsAppBin = require.resolve('create-hops-app');
 
 describe('create-hops-app', () => {
-  const version = 'latest';
+  const version = 'next';
```
ZauberNerd: Should we try to find a way to automatically determine what tag to use here, so that we don't run into the problem again, that we forget to change it after the release?
We could for example:
- look into the package.json and see if the version is a prerelease?
- look into lerna.json, to see if the version is a prerelease?

packages/create-hops-app/package.json
```diff
@@ -11,7 +11,8 @@
     "node": "^10 || ^12"
   },
   "bin": {
-    "create-hops-app": "./index.js"
+    "create-hops-app": "./index.js",
+    "create-hops-app@next": "./index.js"
```
ZauberNerd: One more thing: Last time we forgot to remove `create-hops-app@next` from the bin section after releasing the stable release.
Could we add another test, that reads the bin entries and verifies, that they are correct?

herschel666: I guess there won't be a clean transition path from pre-release to release, but at least we're aware of the issue. 🤷‍♂️

PR #968, opened 2019-10-22
[RFC] add test helper

herschel666:


herschel666:
We decided, that this feature doesn't provide the amount of benefit to the users that would justify the additional maintenance effort that is introduced by it. Hence this PR is closed and we won't further pursue the topic "testing helpers", but will rather focus on improving the documentation with regards to testing.

PR #969, opened 2019-10-23
alias core-js in jest-preset to correct version

ZauberNerd:
Not sure if we should wait for https://github.com/babel/babel/pull/10146
to be released until we release this?

### DIFF
--- a/packages/jest-preset/package.json
+++ b/packages/jest-preset/package.json
@@ -28,6 +28,7 @@
     "@babel/preset-react": "^7.7.0",
     "babel-jest": "^24.9.0",
     "babel-plugin-dynamic-import-node": "^2.3.0",
+    "core-js": "^3.2.1",
     "hops": "^12.0.0-alpha.5",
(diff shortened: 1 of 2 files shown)


ZauberNerd:
We could also investigate whether it makes sense to extract this into a separate babel plugin and replace the webpack aliases in untool with that babel plugin too: https://github.com/untool/untool/blob/next/packages/webpack/lib/configs/build.js#L103-L106

jhiode:
Btw, `v7.7.0` already contains the [fix that you've mentioned](https://github.com/babel/babel/pull/10146): https://github.com/babel/babel/releases/tag/v7.7.0

ZauberNerd:
> Btw, `v7.7.0` already contains the [fix that you've mentioned](https://github.com/babel/babel/pull/10146): https://github.com/babel/babel/releases/tag/v7.7.0

Yes, let's remove it from this PR and set the new minimum required version for babel to v7.7.0

PR #971, opened 2019-10-24
Upgrade untool to 2.0.0-alpha.4

ZauberNerd:


README.md
```diff
@@ -290,6 +291,7 @@ You can provide settings to a Hops application via a `"hops"` key in your `packa
 | `serverDir` | `String` | `node_modules/.cache/untool` | `<rootDir>/dist` | The directory where the generated server middleware will be stored |
 | `browsers` | `Array<String>` | `['defaults']` | `['last 1 Chrome versions']` | An array of browserslist queries to specify targets for which to transpile/polyfill (see [`@babel/preset-env`](https://babeljs.io/docs/en/babel-preset-env) for more information) |
 | `node` | `String` | `current` | `8.10` | A Node.js version identifier or `current` to specify for which target to transpile/polyfill |
+| `browserWhitelist` | `Object` | `{"basePath":true}` | A map of config keys that should be whitelisted and exposed to the client. Nested paths can be described with dot notation |
```
herschel666: Mentioning that those config keys are whitelisted, is meaningless in this context imho.

```diff
- A map of config keys that should be whitelisted and exposed to the client.
+ A map of config keys that should be exposed to the client.
```

`Using` sounds a little better than `with` here imho.

```diff
- Nested paths can be described with dot notation
+ Nested paths can be described using dot notation
```

README.md
```diff
@@ -59,6 +59,7 @@ These are the main features:
     - [Default settings](#default-settings)
     - [Placeholders](#placeholders)
     - [Environment variables](#environment-variables)
+    - [Controlling what will be exposed on the client](#controlling-what-will-be-exposed-on-the-client)
```
herschel666: With the whitelist this is more about `Exposing config values to the client` rather than `Controlling what will be exposed…`, because by default there's nothing exposed besides the `basePath`. So this whole topic isn't about taking care not to expose secrets accidentally anymore, but rather to expose explicitly, what can be exposed.

README.md
```diff
@@ -334,7 +336,7 @@ Usually when you reference `process.env.FOO` in your code, webpack will replace
 
 **But:** Since Hops renders applications on the server- and client-side, we can reference environment variables in the settings and pass them on to the client and therefore allow users to have environment variables that are evaluated at runtime.
 
-**Caution:** Be careful not to expose secrets to the client side, because all environment variables that you reference this way in the settings will be visible in the rendered HTML.
+**Caution:** Be careful not to expose secrets to the client side, because all environment variables that you reference this way in the settings will be visible in the rendered HTML. Read how you can [control what will be exposed on the client](#controlling-what-will-be-exposed-on-the-client).
```
herschel666: This whole paragraph can either be removed.

README.md
```diff
@@ -334,7 +336,7 @@ Usually when you reference `process.env.FOO` in your code, webpack will replace
 
 **But:** Since Hops renders applications on the server- and client-side, we can reference environment variables in the settings and pass them on to the client and therefore allow users to have environment variables that are evaluated at runtime.
```
herschel666: This needs a hint, that exposing those values to the client-side needs an additional effort.

README.md
```diff
@@ -365,6 +367,27 @@ This allows you to have just one build artifact that you can use in different en
 
 **Tip:** Hops ships with integrated support for [`dotenv`](https://github.com/motdotla/dotenv), which means that it will try to read a `.env` file from your application root directory and load its values as environment variables.
 
+#### Controlling what will be exposed on the client
+
+Hops allows you to whitelist specific values from your configuration that will be exposed to the client side. In order to configure which values you need access to on the client side, you can map the configuration keys to boolean values. For nested objects you can use the dot notation:
```
herschel666: Hops doesn't allow, it requires the user to do this additionally.

README.md
```diff
@@ -365,6 +365,27 @@ This allows you to have just one build artifact that you can use in different en
 
 **Tip:** Hops ships with integrated support for [`dotenv`](https://github.com/motdotla/dotenv), which means that it will try to read a `.env` file from your application root directory and load its values as environment variables.
 
+#### Exposing values to the client
+
+Hops requires you to whitelist specific values from your configuration that will be exposed to the client side. In order to configure which values you need access to on the client side, you can map the configuration keys to boolean values. For nested objects you can use the dot notation:
```
jhiode:
```suggestion
Hops requires you to whitelist all values from your configuration that needs to be exposed to the client side. In order to do so, you can map the configuration keys to boolean values. For nested objects you can use the dot notation:
```

We could also write `accessible on the client` instead of `exposed to the client`.

PR #985, opened 2019-11-11
feat: create context extender api

KnisterPeter:
This allows users of the mock server to extend the graphql context by a function.

### DIFF
--- a/packages/apollo-mock-server/context-extender.js
+++ b/packages/apollo-mock-server/context-extender.js
@@ -1,0 +1,1 @@
+module.exports = context => context;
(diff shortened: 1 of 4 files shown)


herschel666:
Can you give an example of what a user would like to put on the context, to make this feature a little more tangible for me? 😃

KnisterPeter:
We would like to use this feature to create a session like handling in the mock-server.
Using the extender pattern allows to decouple the extending code from the mock-server itself.

KnisterPeter:
Since the `extender` is clearly internal API (it should be used by hops devs to extend the graphql context) and it will not be documented further and hops users should not need to use this extension point we decided to not check further on the reponse of the function.

PR #996, opened 2019-12-03
Typescript changes

herschel666:
- strongly recommend to extend `hops-typescript`'s `tsconfig.json`
- check, that the `compilerOptions` `target`- and `moduleResolution`-properties aren't overwritten by the user
- speed up incremental builds by setting `transpileOnly` and `compilerOptions.isolatedModules` to `true` for the dev-server

I'm overall not happy with how the checks are done in the `diagnosis()`-hook & don't lead to an error. Does anyone have a good idea to improve that?

Also the `transpileOnly`-feature doesn't lead to `compilerOptions.isolatedModules` being set to `true` anymore since [this PR](https://github.com/TypeStrong/ts-loader/pull/569), that apparently fixed [this issue](https://github.com/TypeStrong/ts-loader/issues/568). Maybe we don't want to enable it either, but I'm not able to decide this now. 😑

### DIFF
--- a/packages/spec/integration/typescript/package.json
+++ b/packages/spec/integration/typescript/package.json
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "hops": "*",
-    "hops-typescript": "*"
+    "hops-typescript": "*",
+    "typescript": "*"
   }
 }
--- a/packages/typescript/mixin.core.js
+++ b/packages/typescript/mixin.core.js
@@ -5,8 +5,12 @@ const { Mixin } = require('hops-mixin');
 const chalk = require('chalk');
 
 class TypescriptMixin extends Mixin {
-  configureBuild(webpackConfig, { jsLoaderConfig, allLoaderConfigs }) {
+  configureBuild(webpackConfig, { jsLoaderConfig, allLoaderConfigs }, target) {
     const { loader, options, exclude } = jsLoaderConfig;
+    const targetDevelop = target === 'develop';
+    const compilerOptions = targetDevelop
+      ? { compilerOptions: { isolatedModules: true } }
+      : undefined;
 
     allLoaderConfigs.unshift({
       test: /\.tsx?$/,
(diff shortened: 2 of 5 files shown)


packages/spec/integration/typescript/package.json
```diff
@@ -18,5 +18,8 @@
   "dependencies": {
     "hops": "*",
     "hops-typescript": "*"
+  },
+  "devDependencies": {
```
jhiode: It's not a big deal but I think it would be "more correct" do treat `typescript` as a normal dependency.

packages/typescript/mixin.core.js
```diff
@@ -1,10 +1,22 @@
...
+  const { typescript } = { ...dependencies, ...devDependencies };
+
+  return typeof typescript === 'string';
+};
 
 class TypescriptMixin extends Mixin {
-  configureBuild(webpackConfig, { jsLoaderConfig, allLoaderConfigs }) {
+  configureBuild(webpackConfig, { jsLoaderConfig, allLoaderConfigs }, target) {
     const { loader, options, exclude } = jsLoaderConfig;
+    const targetDevelop = target === 'develop';
```
jhiode: Should be skip the webpack configuration if we know that we don't have the typescript dependency installed?
```js
if (!hasTypescriptInstalled(packageJsonPath)) {
  return;
}
```

herschel666: Good point, forgot about that.

herschel666: @jhiode & I agreed that applying this build configuration conditionally makes no sense in the context of `hops-typescript` being the conscious decision of a user to have & use Typescript in their project.

KnisterPeter:
To be honest I'm not sure on how `isolatedModules` behaves in `ts-loader`. I've used it without webpack only.
But I'm really sure that for example parcel is integrating with typescript that way. But it may have some special side effects with webpack.

KnisterPeter:
I would say, lets try it out. If it does work as we hope/expect it to work, then we are a lot faster than without isolated modules.

PR #1000, opened 2019-12-06
fix: upgrade untool to 2.0.0-beta.0

jhiode:
🎉

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -89,7 +89,7 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.6.3", "@babel/core@^7.7.0":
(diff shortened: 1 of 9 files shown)


yarn.lock
```diff
@@ -89,7 +89,7 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.6.3", "@babel/core@^7.7.0":
+"@babel/core@^7.1.0", "@babel/core@^7.7.0":
```
herschel666: why is `@babel/core@^7.1.0` in here?

jhiode: Because of jest-config@^24.9.0:
https://github.com/xing/hops/blob/upgrade-untool/yarn.lock#L7675

PR #1005, opened 2019-12-13
feat(react): warn on non-whitelisted config property

herschel666:
...when it's retrieved in the browser runtime.

### DIFF
--- a/packages/react/config/mixin.runtime.js
+++ b/packages/react/config/mixin.runtime.js
@@ -5,9 +5,33 @@ const { Mixin } = require('hops-mixin');
 
 const { Provider } = require('./context');
 
+const warnOnIncompleteBrowserWhitelist = config => {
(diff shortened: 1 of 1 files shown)


packages/react/config/mixin.runtime.js
```diff
@@ -5,9 +5,33 @@ const { Mixin } = require('hops-mixin');
 
 const { Provider } = require('./context');
 
+const warnOnIncompleteBrowserWhitelist = config => {
+  if (process.env.NODE_ENV === 'development' && typeof Proxy === 'function') {
+    return new Proxy(config, {
+      get: (obj, prop) => {
+        const value = obj[prop];
+
+        if (typeof value === 'undefined') {
```
jhiode:
```suggestion
        if (!(prop in obj)) {
```

PR #1009, opened 2019-12-17
feat(lambda): make node version configurable

herschel666:


packages/lambda/package.json
```diff
@@ -32,5 +31,8 @@
     "serverless-http": "^2.3.0",
     "strip-indent": "^3.0.0"
   },
+  "devDependencies": {
+    "aws-sdk": "^2.546.0"
```
ZauberNerd: How is this supposed to work? The deployment `hops lambda deploy` needs the `aws-sdk` and it won't be installed when it is a peer dependency.

herschel666: My bad. Thought this was ending up in the bundle, which is not the case. Will remove that commit.

package.json
```diff
@@ -49,7 +49,7 @@
       },
       {
         "repository": "https://github.com/foosbyte/kicker-trainer.git",
-        "branch": "next",
+        "branch": "update-node-and-hops",
```
jhiode: Ah wait, we have to remove this first 😬

herschel666: I actually thought about merging it with this change, then merging the kicker-trainer (while keeping the branch) & then revert this change in another PR.

PR #1012, opened 2019-12-20
fix: workaround for css-loader path issue

KnisterPeter:
Since the css-loader changed in the current version to require tilde as prefix for module imports and every other import is taken as relative import, hops follows this and deprecates unprefixed imports.

This in turn conflicts with the postcss-import handler which does not like the tilde prefix and we remove it again here, if its encountered.


KnisterPeter:
The result of this fix is to patch the css AST.
In the current setup, webpack loads the `postcss-loader`, then the `css-loader`.
The postcss is responsible for all processing of imports and css besides css modules 'composes' part. That is done by the `css-loader` which treats all import paths which are not prefixed with '~'  as urls and therefore all paths are relative by default.

Our patch does hook into the `postcss-modules-extract-imports` and rewrites all paths which are not locally resolvable to a path prefixed with '~' which will then be tried to resolve against `node_modules`.

packages/postcss/package.json
```diff
@@ -19,12 +19,15 @@
   },
   "dependencies": {
     "css-loader": "^3.2.0",
+    "core-js": "^3.6.2",
```
ZauberNerd: In https://github.com/xing/hops/blob/master/packages/jest-preset/package.json we have `core-js` as `^3.2.1`, should we align these versions so that they're deduplicated in the yarn.lock?

packages/spec/integration/postcss/index.js
```diff
@@ -1,11 +1,21 @@
...
 import styles from './styles.css';
 /* eslint-disable import/no-unresolved */
 import './global.css?global';
 import 'animate.css/animate.min.css?global';
 /* eslint-enable import/no-unresolved */
 
 export default render(
-  <h1 className={`${styles.headline} fancy-headline animated`}>hello</h1>
+  <>
+    <Helmet>
```
ZauberNerd: Do we still need `Helmet` here? Because the favicon issue only appears when enabling debug mode and it is not directly related to the changes in this PR.

KnisterPeter: I'll move that to another PR

packages/postcss/package.json
KnisterPeter: Go for it 👍

PR #1058, opened 2020-01-31
fix(typescript): use Typescript API for retrieving the final TSConfig

herschel666:


herschel666:
I'm not sure…

KnisterPeter:
What about a test (spec, since it needs to be integration test I think) which checks if a given tsconfig with an extends statement, is really extended?

herschel666:
It's less the conceptional aspect of it rather than the technical one. But I'll find my way around of setting up the mixin in isolation & checking that it's doing its job right. 👍

PR #1062, opened 2020-02-05
chore: use hops-v11.x-branch of foosbyte/kicker-trainer

herschel666:
I created a dedicated Hops-11-branch in `foosbyte/kicker-trainer`. We could then delete the `next`-branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
       },
       {
         "repository": "https://github.com/foosbyte/kicker-trainer.git",
+        "branch": "hops-v11.x",
         "commands": [
           "yarn build -p",
           "yarn test"
--- a/packages/spec/integration/create-hops-app/__tests__/index.spec.js
+++ b/packages/spec/integration/create-hops-app/__tests__/index.spec.js
@@ -5,7 +5,7 @@ const { execSync } = require('child_process');
 const createHopsAppBin = require.resolve('create-hops-app');
 
 describe('create-hops-app', () => {
-  const version = 'latest';
+  const version = '^11';
   const template = 'hops-template-react';
 
   beforeAll(() => {


herschel666:
TODO: revert [63ae067](https://github.com/xing/hops/pull/1062/commits/63ae067ed69d9cce967ac6a70646e8f93c071f4b) when [foosbyte/kicker-trainer#944](https://github.com/foosbyte/kicker-trainer/pull/944) is merged.

ZauberNerd:
You can use `rootManifest.resolutions` to force styled-components v4: https://github.com/xing/canarist#configuration

But also: Do we still need to run canarist tests on v11.x? Because we're not building new features on that branch and if it creates too much hassle on the other projects then maybe it is not worth the effort?

herschel666:
> You can use rootManifest.resolutions to force styled-components v4:

You're right, forgot about that. On the other hand — the PR is pending, so I'd stick with it. Shouldn't make a difference. 🤷‍♂️

> Do we still need to run canarist tests on v11.x?

As long as we're in maintenance mode internally, I think it makes sense to keep on verifying that changes don't bork anything.

ZauberNerd:
> > You can use rootManifest.resolutions to force styled-components v4:
> 
> You're right, forgot about that. On the other hand — the PR is pending, so I'd stick with it. Shouldn't make a difference. 

I meant the downgrade of `styled-components`: https://github.com/foosbyte/kicker-trainer/pull/944 I would rather not merge that but instead use `rootManifest.resolutions` in this repository to force the trainer to use v4.

herschel666:
> I would rather not merge that but instead use rootManifest.resolutions in this repository to force the trainer to use v4.

Why? It's a non-used branch in the kicker-trainer & the situation is imho more RL than brute-forcing down styled-components by the `resolutions`-prop.

ZauberNerd:
Oh, wow. I should have read the PR more carefully. I did not notice, that it targets the v11 branch. I thought you were downgrading on master :D

PR #1063, opened 2020-02-06
feat: support latest Node.js (v13)

ZauberNerd:


ZauberNerd:
Needs a new release of Untool@2 first.

herschel666:
As far as I can assess it, the `create-hops-app`-test isn't fixable atm, because it fetches Hops packages from the registry, that still have the old Node version range, leading to the incompatibility error we see in the test output. Do you agree, @KnisterPeter & @ZauberNerd?

herschel666:
Waiting for #1067

PR #1078, opened 2020-02-18
fix documentation table syntax

nighto:

## Checklist


herschel666:
Hi @nighto, thanks for your contribution! But the `|` in there is correct, because the value can either be of type `Boolean` or of type `Object`. But it needs to be escaped, so that it doesn't get rendered as a column delimiter.

Also please make the PR against the `master`-branch. [We'll take care of cherry-picking it into the `v12.x`-branch after it has been merged.](https://github.com/xing/hops/blob/master/CONTRIBUTING.md#branching-in-hops)

Cheers.

nighto:
Thanks, will close this and propose against master.

Issue #1080, opened 2020-02-18
jest-preset-hops breaks on `importComponent` when using `--coverage`


ZauberNerd:
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

When testing code that uses `importComponent` and running jest with `--coverage` our babel plugin (https://github.com/xing/hops/blob/master/packages/jest-preset/helpers/babel-plugin-import-component.js) breaks, because it runs into instrumented code.

## Proposed solution

Make our babel plugin aware of instrumented code, so that it is able to find the `importComponent` call expression.

PR #1084, opened 2020-02-20
fix(jest-preset): correctly handle coverage instrumented importComponent

ZauberNerd:
When executing jest with `jest --coverage` it adds coverage instrumentation
code to the source code which breaks our babel plugin for the
`importComponent` function. With this change we can correctly handle
instrumented code.

Closes: #1080

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ package-lock.json
 .vscode
 .idea
 .tmp
+coverage/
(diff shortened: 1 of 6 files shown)


herschel666:
Looks good overall & runs on my machine. I lack the Babel plugin API knowledge though, to tell whether this is all fine or not. 😬

PR #1085, opened 2020-02-20
fix: all test files need to run through babel

KnisterPeter:
Since hops depends on a few babel-plugins (e.g. for importComponent) we
need babel to process all files, even in the case of unit tests.

This PR implements the required setup of ts-jest to include the babel
config for hops. This declares that ts-jest should babel all files,
after transpiling with typescript.


herschel666:
It would be nice to have the extraction of `pluginRenameCoreJS` into its own file in a seperate `refactor`-commit. It's hard to tell right now whether the code just shifted or whether there are changes in that function, that need reviewing.

ZauberNerd:
@herschel666 updated. Please take another look.

PR #1106, opened 2020-03-09
fix(react): ignore prop-types symbol check

herschel666:
...in config proxy, to prevent warnings concerning a missing
"@@toStringTag"-property.


ZauberNerd:
bors merge

PR #1125, opened 2020-04-01

jhiode:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 cache:
-  yarn: true
+  yarn: false
   directories:
     - node_modules
 


herschel666:
bors merge

herschel666:
How to convey to bors, that we don't really care about `canarist/0` atm?? /cc @ZauberNerd @KnisterPeter

KnisterPeter:
@herschel666  The canarist failed (#3373). What is the reason here?

herschel666:
@KnisterPeter As I stated above, this is a known error we want to tackle later & thus ignore for now. The details are written down in SOLAR-628.

ZauberNerd:
@herschel666 I would suggest to either fix the test or disable it in the canarist service. Bors does here what it should do and that is to block a merge if any of the checks fails.

herschel666:
@ZauberNerd Sounds legit! 👍

herschel666:
bors merge

herschel666:
bors merge

herschel666:
bors merge

PR #1126, opened 2020-04-01

jhiode:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
     - HOST=0.0.0.0
 
 cache:
-  yarn: true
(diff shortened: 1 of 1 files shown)


herschel666:
We have a different issue here with the Canarist run against internal Hops. 😢

herschel666:
Addendum: in `#3367` the change is tested against the `master`-branch of internal Hops, which is wrong. @ZauberNerd What's the current state of detecting the base branch in the Canarist Service?

herschel666:
Since everything seems to be fine with Travis atm, I dare to close this PR. Feel free to re-open, if you object.

PR #1152, opened 2020-04-20
refactor: integrate untool into Hops

KnisterPeter:


DOCUMENTATION.md
```diff
@@ -548,7 +547,7 @@ import { render, importComponent } from 'hops';
 
 const Home = importComponent(
   () => import('./home'),
-  namespace => namespace.Home
+  (namespace) => namespace.Home
```
matthias-reis, line 550: Looks like the new prettier version :-)

KnisterPeter: 2.0 :wink:

packages/apollo-mock-server/mixin.core.js
```diff
@@ -1,7 +1,7 @@
 const { Mixin } = require('hops-mixin');
 const {
   internal: { createWebpackMiddleware, StatsFilePlugin },
-} = require('@untool/webpack');
+} = require('hops-webpack');
```
matthias-reis, line 4: Would it make sense to think of a namespace? (Don't remember whether `@hops` still was available)

PR #1206, opened 2020-05-12
fix(react): render react-helmet-async values on the server again

bxt:
This pull request closes issue # (_no issue yet_)

## Current state

As specified in the [documentation of `react-helmet-async`](https://github.com/staylor/react-helmet-async#usage) it will create a property `helmet` in the object passed as context. Back when the [react package from untool](https://github.com/untool/untool/blob/ef7abccc693ee5fb24c1c17bc7a33098aafed0b1/packages/react/mixins/mixin.server.js#L47) was used, this context was saved in a context variable and later its `helmet` key was used. Now the context is saved in a variable `helmet`, but the nested `helmet` key of this context is never taken out. So instead of adding keys like `title`, `base`, `meta`, etc. to the fragments so they can be rendered by the template, only a key called `helmet` is added which the template silently ignores. This means everything set by helmet gets lost on the server. Luckily the client fixes that almost immediately after the JavaScript starts running in the browser, so the problem went unnoticed for some time.

## Changes introduced here

This adds back the concept of a context and then properly takes out the helmet key form it, so that the individual entries are copied to the fragments again and thus get rendered in the server-generated HTML response.

It adds a basic integration test to see that it actually works and to avoid regressions in the future.

Unfortunately the extraction of the `helmet` key had to be moved to the call-site, because it does not exist yet when `bootstrap` is called.

## Checklist

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/react/helmet/mixin.runtime.js
+++ b/packages/react/helmet/mixin.runtime.js
@@ -5,17 +5,17 @@ const { HelmetProvider } = require('react-helmet-async');
 class ReactHelmetMixin extends Mixin {
   constructor(config, _element, options) {
     super(config, options);
-    this.helmet = {};
+    this.context = {};
   }
 
   bootstrap(_req, res) {
     if (res) {
-      res.locals.helmet = this.helmet;
+      res.locals.helmetContext = this.context;
     }
   }
 
   enhanceElement(element) {
-    return createElement(HelmetProvider, { context: this.helmet }, element);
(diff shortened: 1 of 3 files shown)


ZauberNerd:
bors try

herschel666:
bors retry

ZauberNerd:
bors retry

ZauberNerd:
bors merge

PR #1207, opened 2020-05-13
fix(react): render react-helmet-async values on the server

ZauberNerd:
As specified in the documentation of react-helmet-async[1] it will create
a property helmet in the passed context. Back when the react package from
untool[2] was used, this was saved in a context variable and later taken out.
Now the context is saved in a helmet variable, but the nested helmet key is
never taken out. So instead of adding keys like title, base, meta, etc. to
the fragments, only a key called helmet was added, so everything set by helmet
got lost on the server. Luckily the client immediately fixes that, so the problem
went unnoticed for some time.

This adds back the context concept and then properly takes out the helmet
key form it, so that the individual entries are copied to the fragments again.

Also it adds an integration test to see that it actually works and to avoid
regressions in the future.

  1: https://github.com/staylor/react-helmet-async#usage
  2: https://github.com/untool/untool/blob/ef7abccc693ee5fb24c1c17bc7a33098aafed0b1/packages/react/mixins/mixin.server.js#L47

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/react/helmet/mixin.runtime.js
+++ b/packages/react/helmet/mixin.runtime.js
@@ -5,17 +5,17 @@ const { HelmetProvider } = require('react-helmet-async');
 class ReactHelmetMixin extends Mixin {
   constructor(config, _element, options) {
     super(config, options);
-    this.helmet = {};
+    this.context = {};
   }
 
   bootstrap(_req, res) {
     if (res) {
-      res.locals.helmet = this.helmet;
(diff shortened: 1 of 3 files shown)


ZauberNerd:
bors try

ZauberNerd:
bors merge

PR #1228, opened 2020-05-26
feat(webpack): print total bundle size

jhiode:
example output:
```
hops-demo:info bundling 'node' finished after 30.7s (12.4 MB)
hops-demo:info bundling 'build' finished after 51.2s (7.34 MB)
```

### DIFF
--- a/packages/webpack/lib/plugins/log.js
+++ b/packages/webpack/lib/plugins/log.js
@@ -38,7 +38,13 @@ const formatWarning = (name, duration, assets, isRebuild) => {
 };
 
 const formatSuccess = (name, duration, assets, isRebuild) => {
-  const message = `bundling '${name}' finished after ${duration}`;
+  const totalSize = Object.values(assets).reduce(
+    (sum, asset) => sum + asset.size,
+    0
+  );
+  const message = `bundling '${name}' finished after ${duration} (${prettyBytes(
+    totalSize
+  )})`;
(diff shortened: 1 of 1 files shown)


packages/webpack/lib/plugins/log.js
```diff
@@ -38,7 +38,13 @@ const formatWarning = (name, duration, assets, isRebuild) => {
 };
 
 const formatSuccess = (name, duration, assets, isRebuild) => {
-  const message = `bundling '${name}' finished after ${duration}`;
+  const totalSize = Object.values(assets).reduce(
+    (sum, asset) => sum + asset.size,
+    0
+  );
+  const message = `bundling '${name}' finished after ${duration} (${prettyBytes(
```
ZauberNerd, line 45: I would probably print here something like `bundling .. finished after .. combined size of all assets: ...`

jhiode: I just used the same format as the other assets:
```
hops-demo:info bundling 'node' finished after 1m 5.1s (14.7 MB)
- server.js (14.5 MB)
- stats.json (187 kB)
```

ZauberNerd: Ok, let's leave it like that then. I just thought it might be irritating if there's no explanation what this number means.

jhiode:
bors merge

PR #1229, opened 2020-05-26
feat(webpack): print total bundle size

jhiode:
Same as #1228


herschel666:
bors retry

herschel666:
bors retry

PR #1230, opened 2020-05-27
feat(webpack): introduce experimental fast-build mode

jhiode:
The new experimental `--fast-build` feature only transpiles some node_modules instead of all.
That also requires to change core-js `useBuiltIns` to `entry` as we don't know which polyfills we need in the end.

**`hops build -p`**
```
fixture-hops:info bundling 'node' finished after 8.4s (2.92 MB)
- server.js (2.89 MB)
- stats.json (33 kB)
fixture-hops:info bundling 'build' finished after 8.4s (1.69 MB)
- fixture-hops-0-1b0f62f1564b.js (255 B)
- fixture-hops-2-ef56cba710d4.js (338 kB)
- fixture-hops-ad490021bce4.js (39.5 kB)
```

**`hops build -p --fast-build`**
```
fixture-hops:info bundling 'node' finished after 3.2s (2.56 MB)
- server.js (2.53 MB)
- stats.json (33 kB)
fixture-hops:info bundling 'build' finished after 3.8s (1.68 MB)
- fixture-hops-0-1b0f62f1564b.js (255 B)
- fixture-hops-06324ff963f7.js (37.7 kB)
- fixture-hops-2-ba3747322e3f.js (331 kB)
```

### DIFF
--- a/packages/webpack/lib/shims/node.js
+++ b/packages/webpack/lib/shims/node.js
@@ -2,10 +2,6 @@ if (!require('module').prototype._compile.__sourceMapSupport) {
   require('source-map-support/register');
 }
 
-if (process.env.__HOPS_FAST_DEV__ !== 'true') {
-  require('core-js');
-}
-
 if (module.hot) {
   require('webpack/hot/log').setLogLevel('none');
   module.hot.accept('hops/entrypoint');
--- a/packages/webpack/mixins/optimizations/mixin.core.js
+++ b/packages/webpack/mixins/optimizations/mixin.core.js
@@ -11,6 +11,10 @@ function include(patterns) {
     patterns.some((rule) => minimatch(path, rule));
 }
 
+function unique(value, index, self) {
+  return self.indexOf(value) === index;
+}
+
 function findResolved(a, b) {
   try {
     return a === require.resolve(b);
@@ -21,17 +25,18 @@ function findResolved(a, b) {
 
 class WebpackOptimizationsMixin extends Mixin {
   configureBuild(webpackConfig, { allLoaderConfigs, jsLoaderConfig }, target) {
-    const { fastDev } = this.options;
+    const { fastBuild, fastDev } = this.options;
 
-    if (!fastDev) {
+    if (!fastDev && !fastBuild) {
       return;
     }
 
     const { NODE_ENV } = process.env;
     const isDevelop = target === 'develop' || NODE_ENV !== 'production';
     const skipPolyfilling = fastDev && isDevelop;
+    const skipNodeModules = !skipPolyfilling && fastBuild;
 
-    if (skipPolyfilling) {
+    if (skipPolyfilling || skipNodeModules) {
       const { experimental: { babelIncludePatterns = [] } = {} } = this.config;
 
       jsLoaderConfig.include = include(babelIncludePatterns);
@@ -47,17 +52,30 @@ class WebpackOptimizationsMixin extends Mixin {
           .find(([path]) => findResolved(path, '@babel/preset-env')) || [];
 
       if (presetEnvOptions) {
-        presetEnvOptions.useBuiltIns = false;
-        delete presetEnvOptions.corejs;
-      }
+        if (skipPolyfilling) {
+          presetEnvOptions.useBuiltIns = false;
+          delete presetEnvOptions.corejs;
+        } else if (skipNodeModules) {
+          presetEnvOptions.useBuiltIns = 'entry';
 
+          webpackConfig.entry = [
+            require.resolve('core-js/stable'),
+            require.resolve('regenerator-runtime/runtime'),
+          ]
+            .concat(webpackConfig.entry)
+            .filter(unique);
+        }
+      }
+    }
+
+    if (skipPolyfilling) {
       allLoaderConfigs
         .filter((loader) => {
           return loader.test instanceof RegExp && loader.test.test('.tsx');
         })
         .forEach((loader) => {
-          const babelLoader = loader.use.find((l) => {
-            return findResolved(l.loader, 'babel-loader');
+          const babelLoader = loader.use.find(({ loader }) => {
+            return findResolved(loader, 'babel-loader');
           });
 
           if (babelLoader) {
--- a/packages/webpack/mixins/optimizations/mixin.core.js
+++ b/packages/webpack/mixins/optimizations/mixin.core.js
@@ -1,4 +1,15 @@
+const minimatch = require('minimatch');
 const { Mixin } = require('hops-bootstrap');
+
+function include(patterns) {
+  return (path) =>
+    // allow everything outside of node_modules
+    !path.includes('node_modules') ||
+    // allow all hops packages
+    /node_modules[/\\]hops/.test(path) ||
+    // check if any other rule matches
+    patterns.some((rule) => minimatch(path, rule));
+}
 
 function findResolved(a, b) {
   try {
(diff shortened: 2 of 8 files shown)


packages/webpack/lib/shims/develop.js
```diff
@@ -1,5 +1,13 @@
 /* eslint-env browser */
 
+if (process.env.__HOPS_FAST_BUILD__ === 'true') {
+  if (process.env.__HOPS_FAST_DEV__ !== 'true') {
```
KnisterPeter, line 4: Can you add a comment here, why we need to disable this with fast-dev?

KnisterPeter: Thats probably, because we do not polyfill at all with fast-dev right?
But I think its not obvious in this case

jhiode: Yes exactly, I can also leave a comment in code 👍

packages/webpack/mixins/fast-build/mixin.core.js
```diff
@@ -0,0 +1,45 @@
...
+      return;
+    }
+
+    process.env.__HOPS_FAST_BUILD__ = 'true';
+
+    const {
+      experimental: { fastBuild: { includeRules = [] } = {} } = {},
+    } = this.config;
+    const rules = includeRules.map((rule) => new RegExp(rule));
+    const options = jsLoaderConfig.options.presets[0][1];
```
herschel666, line 30: I'd use `[].find(…)` here to be on the save side. You'll still be able the mutate the preset's options object.

packages/webpack/mixins/fast-build/mixin.core.js
```diff
@@ -0,0 +1,45 @@
...
+      this.options.fastBuild || (this.options.fastDev && target === 'develop');
+
+    if (!enabled) {
+      return;
+    }
+
+    process.env.__HOPS_FAST_BUILD__ = 'true';
+
+    const {
+      experimental: { fastBuild: { includeRules = [] } = {} } = {},
```
herschel666, line 27: here they're named `includeRules`, down in the `preset.js` they're named `buildIncludeRules`. however, I think I'd prefer `babelIncludeRules`, because there's a lot to in- & exclude when dealing with webpack. :D

packages/webpack/lib/shims/build.js
```diff
@@ -1,8 +1,10 @@
...
-import entryPoint from 'hops/entrypoint';
-/* eslint-enable node/no-missing-import */
-/* eslint-enable import/no-unresolved */
+if (process.env.__HOPS_FAST_BUILD__ === 'true') {
+  require('core-js/stable');
+  require('regenerator-runtime/runtime');
+}
+
+/* eslint-disable-next-line node/no-missing-require */
+const entryPoint = require('hops/entrypoint').default;
```
ZauberNerd, line 7: I would like to keep using ESM in the build shim because that makes tree shaking easier. For dev it is fine, because we don't have tree shaking there.
Maybe we could split the shim into two build shims or use the array syntax of webpack entry points (https://webpack.js.org/concepts/entry-points/#single-entry-shorthand-syntax), like:
```js
entry: [require.resolve('core-js/stable'), require.resolve('regenerator-runtime/runtime'), 'require.resolve('../shims/build')],
```

packages/webpack/lib/shims/node.js
```diff
@@ -3,7 +3,11 @@ if (!require('module').prototype._compile.__sourceMapSupport) {
 }
 
 if (process.env.__HOPS_FAST_DEV__ !== 'true') {
-  require('core-js');
+  require('core-js/stable');
```
ZauberNerd, line 6: :+1:

packages/webpack/mixins/fast-build/mixin.core.js
```diff
@@ -0,0 +1,45 @@
...
+    // check if any other rule matches
+    rules.some((rule) => rule.test(path));
+
+  return shouldInclude;
+}
+
+class HopsFastBuildMixin extends Mixin {
+  configureBuild(webpackConfig, { jsLoaderConfig }, target) {
+    const enabled =
+      this.options.fastBuild || (this.options.fastDev && target === 'develop');
```
ZauberNerd, line 18: Why do we set `enabled` when we're in `fastDev` mode?

jhiode: Because dev mode would also benefit from less files to transpile. But `target === 'develop'` is actually wrong, we should also do this for `target === 'node'`.

packages/webpack/mixins/fast-build/mixin.core.js
jhiode: Ok, I have changed that already but lets wait for more feedback first.

```js
    const [, options = {}] =
      jsLoaderConfig.options.presets.find(([path]) =>
        /@babel[/\\]preset-env/.test(path)
      ) || [];
```

packages/webpack/mixins/fast-build/mixin.core.js
ZauberNerd: I'm still not sure, why we need to handle fast-dev here too, because:
- each mixin should be encapsulated and care only about its own config
- fast-dev and fast-build can never happen at the same time, because fast-dev has an exit clause for target=build.

And if there's something that would also benefit fast-dev we should include it in fast-dev, I think?
Also: We're already excluding all of node_modules in fast-dev: https://github.com/xing/hops/blob/master/packages/webpack/mixins/fast-dev/mixin.core.js#L23

jhiode: > each mixin should be encapsulated and care only about its own config
Sure, but if I would name this `babel-transpile-whitelist` it would be related to `--fast-dev`.
So the issue is more about names than the actual concerns I think.

> fast-dev and fast-build can never happen at the same time, because fast-dev has an exit clause for target=build.

Not for the develop target but we have two others (mock-server and node) that would still benefit from these chages.

I mean we could think about decoupling the flags (fast-dev and fast-build) from the optimizations and just pick from a list of optimizations for each mode...

package.json
```diff
@@ -27,6 +27,7 @@
   ],
   "scripts": {
     "preinstall": "node -e \"process.exit(process.env.npm_execpath.includes('yarn') ? 0 : 1)\"",
+    "hops": "cd packages/template-react; yarn hops",
```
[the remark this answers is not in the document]
jhiode: No, why? This allows me to run any hops command from the root. Or do you ask me to create a separate PR for this?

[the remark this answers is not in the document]
herschel666: I like the idea, but i'd rather use a project in spec/integrations for that. So maybe we should tackle this later in a different PR.

packages/webpack/mixins/optimizations/mixin.core.js
```diff
@@ -0,0 +1,102 @@
...
+          presetEnvOptions.useBuiltIns = false;
+          delete presetEnvOptions.corejs;
+        } else if (optimizeCoreJsEntry) {
+          presetEnvOptions.useBuiltIns = 'entry';
+
+          webpackConfig.entry = [
+            require.resolve('core-js/stable'),
+            require.resolve('regenerator-runtime/runtime'),
+          ]
+            .concat(webpackConfig.entry)
```
herschel666, line 67: Why do you use the spread-syntax in objects, but not in arrays? 😜

packages/webpack/mixins/optimizations/mixin.core.js
```diff
@@ -0,0 +1,102 @@
...
+  configureBuild(webpackConfig, { allLoaderConfigs, jsLoaderConfig }, target) {
+    const { fastBuild, fastDev } = this.options;
+
+    if (!fastDev || !fastBuild) {
+      return;
+    }
+
+    const { NODE_ENV } = process.env;
+    const isDevelop = target === 'develop' || NODE_ENV !== 'production';
+    const optimizeNoCoreJs = fastDev && isDevelop;
```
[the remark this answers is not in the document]
ZauberNerd: I like `skipPolyfilling` and `skipNodeModules` - we're not not bundling them, we're just not transpiling them.

packages/webpack/mixins/optimizations/mixin.core.js
jhiode: Because `webpackConfig.entry` isn't always an array.

PR #1233, opened 2020-05-28
Revert "feat(webpack): improve feedback on conflicting core-js versions"

herschel666:
Reverts xing/hops#1224

It's not been the right approach unfortunately. The issue isn't per sé that there are multiple major versions around. The issue's rather that a dependency might import from `core-js` & thereby expects a different major version, which leads to errors, because the structure of the deep imports has changed.

So the way to go is rather to listen for those errors & tell the user, that this specific dependency is doing it wrong.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,6 +1,6 @@
 node_modules/
-packages/spec/*/**/build
-packages/spec/*/**/dist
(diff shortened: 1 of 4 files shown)


herschel666:
bors merge

Issue #1257, opened 2020-06-20
Module resolution fails in symlinked preset/mixin package 📦, and does not abide to "--preserve-symlinks" node option


hossam-magdy:
## This is my intent (choose one)

- [x] I want to report a bug
- [x] I want to request a feature or change
- [x] I want to provide or change a feature

## The problem

While developing a mixin-and-preset package to be loaded inside any hops application, using `yarn link`, `npm link` or "yarn workspaces", the package is not hooked gracefully during the bootstrap phase of hops.

In other words, resolving peerDependencies of a **symlinked** hops mixin-and-preset package fails, even when using the [`--preserve-symlinks` node arg](https://nodejs.org/api/all.html#cli_preserve_symlinks) which as describe in the official node documentation as a way to interpret node resolution algorithm to treat symlinks as preserved (as if they are normal dirs). A further detailed article can be found [here](https://chevtek.io/you-can-finally-npm-link-packages-that-contain-peer-dependencies/).

### Steps done to reproduce:

Although it was not that straight-forward to get it concrete, but I think the steps to reproduce the issue, can be narrowed down to the following:

1. Have/Create any working/runnable hops application. Assuming in path `~/hops-app`.
2. Create a new separate directory, assuming in path `~/my-hops-preset`, with:

   ```js
   // file: ~/my-hops-preset/package.json
   // including a `preset` field, as stated in docs: https://github.com/xing/hops/blob/v12.x/DOCUMENTATION.md#presets-1
   {
     "name": "my-hops-preset",
     "preset": "path-to/preset.js",
     "peerDependencies": {
       "hops": "*",  // or "hops-mixin"
     },
   }
   ```

   ```js
   // file: ~/my-hops-preset/path-to/preset.js
   module.exports = { mixins: [__dirname] };
   ```

   ```js
   // file: ~/my-hops-preset/path-to/mixin.core.js
   // As described in docs: https://github.com/xing/hops/blob/v12.x/DOCUMENTATION.md#creating-your-first-core-mixin
   const { Mixin } = require('hops');

   class MyMixin extends Mixin {
     configureBuild(webpackConfig) {
       webpackConfig.resolve.alias['my-modules'] = path.resolve(
         this.config.rootDir,
         'my-aliased-modules'
       );
     }
   }

   module.exports = MyMixin;
   ```

3. Add `my-hops-preset` as a dependency (or devDep) in the `hops-app`

   ```js
   // file: ~/hops-app/package.json
   {
     "dependencies": {
[...]

PR #1258, opened 2020-06-20
fix: module resolution in symlinked package of a preset

hossam-magdy:
This pull request closes issue https://github.com/xing/hops/issues/1257

**Detailed explanation and analysis can be found in [the related issue](https://github.com/xing/hops/issues/1257)**

## Current state

As described in [the issue](https://github.com/xing/hops/issues/1257):

Resolving peerDependencies of a **symlinked** hops mixin-and-preset package fails, even when using the [`--preserve-symlinks` node arg](https://nodejs.org/api/all.html#cli_preserve_symlinks) which as describe in the official node documentation as a way to interpret node resolution algorithm to treat symlinks as preserved (as if they are normal dirs). Further detailed article can be found [here](https://chevtek.io/you-can-finally-npm-link-packages-that-contain-peer-dependencies/).

## Changes introduced here

In [the resolver of bootstrap package](https://github.com/xing/hops/blob/e7a3b46/packages/bootstrap/lib/resolver.js), during any usage of [`createResolver`](https://github.com/webpack/enhanced-resolve#creating-a-resolver) we need to pass `symlinks` option as `false` if the env `NODE_PRESERVE_SYMLINK=1`.

It was proven (locally) that loading the hops-preset and hops-mixins work correctly and as expected by adding `symlinks: false` option to the createResolver (alongside env:`NODE_PRESERVE_SYMLINK=1`), specifically in [resolvePreset](https://github.com/xing/hops/blob/e7a3b46/packages/bootstrap/lib/resolver.js#L12) and [resolveMixins](https://github.com/xing/hops/blob/e7a3b46/packages/bootstrap/lib/resolver.js#L17).

**UPDATE**: I updated this PR, initially to handle the node args via `process.execArgv` (thanks to @ZauberNerd for pointing me out to it), but then also updated to set `symlinks` option to be always `false`, which will solve the issue. Aside to mention: after further investigation, the initial/main pinpoint of the issue was [the resolution of presetJs file](https://github.com/xing/hops/pull/1258/commits/34ef9afeab857144165a9fe30c5d35dbb56c450f#diff-7a1d45a005c82723268a0604aa23a9d6R18) (which is why I updated those few line only for clear names that helped me during debugging). If the presetJs file is resolved to the real path, the mixins in the config object are of real path, and any package loaded within those mixins will be resolved relative to the real path. Yet the `symlink:false` is needed for both the preset and mixin resolution.

## Checklist

- [ ] <del>Necessary unit tests are added in order to ensure correct behavior</del> (I am not sure how this could be possibly tested. If any of the reviewers has any idea, I would be curious to know!)
- [ ] <del>Documentation has been added</del> (Is it applicable for such change?!)

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/bootstrap/lib/loader.js
+++ b/packages/bootstrap/lib/loader.js
@@ -15,7 +15,9 @@ class Loader {
   }
   load(context, module) {
     const { load } = cosmiconfig(this.namespace);
-    return load(resolvePreset(context, module));
+    const presetJsFilePath = resolvePreset(context, module);
+    const loadedPreset = load(presetJsFilePath);
+    return loadedPreset;
   }
   search(stopDir) {
     const { search } = cosmiconfig(this.namespace, { stopDir });
--- a/packages/bootstrap/lib/resolver.js
+++ b/packages/bootstrap/lib/resolver.js
@@ -7,14 +7,7 @@ const {
   create: { sync: createResolver },
 } = require('enhanced-resolve');
 
-const defaultConfig = {
-  symlinks: !(
-    process.env.NODE_PRESERVE_SYMLINKS == 1 ||
-    (typeof process.env.NODE_OPTIONS === 'string' &&
-      process.env.NODE_OPTIONS.includes('--preserve-symlinks')) ||
-    process.execArgv.includes('--preserve-symlinks')
-  ),
-};
+const defaultConfig = { symlinks: false };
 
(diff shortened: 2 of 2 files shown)


ZauberNerd:
bors try

PR #1315, opened 2020-08-13
fix(react-apollo): prevent requests on server when SSR is turned off

herschel666:
The used `ApolloLink.empty()` doesn't seem to be a part of the official public API. But we're having a test in place, so we should™ be notified when it ceases to exist.

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


herschel666:
Actually wanted to open this as a draft PR… 🙄

ZauberNerd:
You can convert it to a draft:
draft

herschel666:
Okay, I found something over at Apollo's: [Using a valid ApolloLink with renderToString executes queries on the server side](https://github.com/apollographql/apollo-client/issues/6681#issuecomment-663166311).

Setting the network policies to `cache-only` I already tried without success. So I'd then dump the misused `SchemaLink` and set up some kind of dummy link as described in the comment.

packages/spec/integration/graphql-no-ssr/__tests__/develop.js
```diff
@@ -0,0 +1,16 @@
...
+
+  beforeAll(async () => {
+    url = await HopsCLI.start('--fast-dev');
+  });
+
+  it('renders the page correctly', async () => {
+    const { page, getInnerText } = await createPage();
+    await page.goto(url, { waitUntil: 'networkidle2' });
+
+    expect(await getInnerText('p')).toBe('Hello World!');
```
ZauberNerd, line 12: Shouldn't we also assert that:
* nothing was rendered server-side
* no fetch requests have been dispatched on the server?

herschel666: Yeah, it should check for `Loading...`. Will add that.

The query-part is actually done, but a little disguised. Only the first query will receive the `subject` `"World"`. All subsequent queries receive `"SSR"`. So the test will fail, when there has been a server-side query prior to the one on the client.

herschel666: Okay, switched the wording to `Client` and `Server`. So when the text of the rendered app says `Hello Server!`, something went wrong, a request has been made on the server & the test will notify us.

packages/spec/integration/graphql-no-ssr/mock-schema.js
```diff
@@ -0,0 +1,34 @@
...
+    id: ID!
+    title: String
+  }
+
+  type Query {
+    subject: Subject
+  }
+`;
+
+const values = Array.from({ length: 10 }, () => ({
```
ZauberNerd, line 14: Why do we need the `Server` value 10 times? Couldn't we just write it out?

herschel666: For debugging purposes when you run this app and check it in the browser it's less annnoying, when you can reload the page more than once, before an error is thrown because of the GraphQL API not responding with a string value. 🤷‍♂️

packages/spec/integration/graphql-no-ssr/mock-schema.js
```diff
@@ -0,0 +1,34 @@
...
+  id: 'def.456',
+  title: 'Server',
+})).concat({
+  id: 'abc.123',
+  title: 'Client',
+});
+
+const resolvers = {
+  Query: {
+    subject: () => values.pop(),
```
jhiode, line 24: How about something that is easier to understand and also always returns something without an error? 🙂

```suggestion
    subject: () => {
      id: 'abc.123',
      title: `Fetch no. #${++number}`,
    },
```

herschel666:
bors merge

PR #1317, opened 2020-08-20
feat: allow mixin instantiation without side-effects

ZauberNerd:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/bootstrap/index.js
+++ b/packages/bootstrap/index.js
@@ -11,7 +11,32 @@ exports.initialize = function initialize(overrides = {}, ...args) {
   return exports.bootstrap(config, mixins, ...args);
 };
 
-exports.bootstrap = function bootstrap(config, mixins, ...args) {
+function hasClasses(mixins) {
+  return mixins.every((mixin) => mixin instanceof Object);
+}
+
+exports.bootstrap = function bootstrap(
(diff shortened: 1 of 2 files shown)


packages/bootstrap/index.js
```diff
@@ -14,9 +14,32 @@ exports.Mixin = class Mixin {
   }
 };
 
-exports.initialize = function initialize(overrides = {}, ...args) {
-  const config = getConfig(overrides);
-  const mixins = getMixins(config);
+function hasClasses(mixins) {
+  return mixins.every((mixin) => mixin instanceof Object);
```
KnisterPeter, line 18: Why is this called `hasClasses` if the implementation means `hasObjects`?

jhiode (commented):
This is already a big step in the right direction and I am fine in doing this as the first step.

I would still love to keep the bootstrap function from the previous commit. For that to work, we also need to export the `environmentalize` and `placeholdify` functions. And `environmentalize` needs to accept a custom env object (it could have a fallback though). We could combine both to `processConfig` in order to make it simple to use.

Example:
```js
bootstrap(processConfig(hopsConfig, { NODE_ENV: 'foo' }), mixins);
```

PR #1323, opened 2020-08-28
fix(bootstrap): re-export strategies in hops-bootstrap

jhiode:
Otherwise, we would need to add mixinable to all internal Hops packages which would
be unfortunate and also makes it possible to have multiple versions of mixinable in
the app if the versions diverge.

The deprecation got introduced in #1162 and is only available in v13 AFAICT.

### DIFF
--- a/DEPRECATIONS.md
+++ b/DEPRECATIONS.md
@@ -14,4 +14,4 @@ Deep imports to `hops-react/lib/runtime` are deprecated and should be changed to
 
 ### DEP0005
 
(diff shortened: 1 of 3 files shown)


herschel666 (changes requested):
The [README](https://github.com/xing/hops/tree/ebe4de5e09b2ddf883d92c7da690fe8f33f8bb83/packages/bootstrap#api) needs an update.

herschel666 (approved):

jhiode:
bors merge

PR #1340, opened 2020-09-17
fix(webpack): use webpack HMR for node build to fix memory leak

ZauberNerd:
This was implemented previously already but got lost during the
implementation of another feature which was intended to reduce memory
consumption.

We are using webpack hot module reload for the server bundle, to be able
to update a running server with incremental changes.
In order to use the HMR on the server side, we need to send a signal to
the process which tells webpack to reloaad the module.
This is the part that got lost and instead we re-required the server
bundle every time a watch event happened on the webpack compiler.

In the current implementation we require the server bundle just once and
let webpack HMR handle all other updates.

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/webpack/lib/middlewares/render.js
+++ b/packages/webpack/lib/middlewares/render.js
@@ -17,10 +17,18 @@ function createRenderMiddleware(buildConfigArgs, watch, mixin) {
   }
 
   const enhancedPromise = new EnhancedPromise();
+  let middleware;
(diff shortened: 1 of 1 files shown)


KnisterPeter (approved):

ZauberNerd:
bors merge

PR #1352, opened 2020-09-22
fix(webpack): replace serialize-error with a custom implementation

jhiode:
For some reason, serialize-error is unable to successfully detect and resolve a circular error object. The result is an out-of-memory error when the compilation fails. I haven't tested if this is the case for all errors or only some specific ones though.

Another flaw of the previous implementation is that we lose the information whether the error is a compile or build error as the error deserialization always created a plain `Error` instance.
And as an error object only consists of the [three fields `name`,  `message` and  `stack`](https://github.com/microsoft/TypeScript/blob/7c6462aa10329884618c51ff9af76c66b503f47c/lib/lib.es5.d.ts#L972-L976
) the serialization of any other fields seems unnecessary to me.

### DIFF
--- a/packages/webpack/lib/utils/compiler-fork.js
+++ b/packages/webpack/lib/utils/compiler-fork.js
@@ -1,8 +1,11 @@
 const { join } = require('path');
-const { serializeError } = require('serialize-error');
 
 const { configure } = require('../../');
-const { BuildError, CompilerError } = require('../utils/errors');
+const {
+  BuildError,
+  CompilerError,
+  serializeError,
+} = require('../utils/errors');
 const { createCompiler } = require('./compiler');
 
 process.on('message', (message) => {
--- a/packages/webpack/lib/utils/compiler.js
+++ b/packages/webpack/lib/utils/compiler.js
@@ -3,7 +3,11 @@ const { fork } = require('child_process');
 const Observable = require('zen-observable');
 const sourceMapSupport = require('source-map-support');
 
-const { BuildError, CompilerError } = require('../utils/errors');
+const {
+  BuildError,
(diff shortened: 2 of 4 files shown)


ZauberNerd (approved):

jhiode:
bors merge

herschel666 (commented):
Two nit-picks…

And have you checked whether you somehow end up in this [`if`-block](https://github.com/sindresorhus/serialize-error/blob/3235f503963c45d30bf43547a5e97c210598ded4/index.js#L43) with a circular value?

In general I'd feel more safe to use a widely used OS solution here. And if there's a bug in it, we should contribute a fix. Given we have a fix, of course!

jhiode:
> In general I'd feel more safe to use a widely used OS solution here. And if there's a bug in it, we should contribute a fix. Given we have a fix, of course!

@herschel666 Yes, but as I already wrote I don't see the benefit of serializing all properties. I mean if it's a special error class we would still miss any functions and testing via `instanceof` would also not work.
And the only errors that we are serializing are the two that I test for in the if/else block. Having more properties is in my opinion not worth anything.

Regarding the nit-picks: Fair points but I don't want to create PRs for them to be honest.

jhiode:
But if you disagree feel free to further improve this in a new PR 🙂

herschel666:
@jhiode Okay, I'm pretty sure you know what you're doing! And personally I don't feel the urge to cover those two nit-picks in a separate PR right now (`#priorities`). But I would love to see switch/case for those use-cases in future code changes. 😜

Anyways, you forgot to update the `yarn.lock`-file. This needs to happen. 🤷‍♂️

PR #1357, opened 2020-09-24
Improve diagnose hook

herschel666:
**ToC**

- the `doctor` got two new methods: `pushWarning(message: string)` and `pushError(tag: string, message: string)`; errors are completely listed, the last one ends the process with exit code `1`
- the `diagnose`-hook got a second argument: `mode: string`, which is either `build`, `develop`, `serve`, `build-serve` or `indeterminate`.
- errors can be converted into non-fatal warnings, by passing their tag via `HOPS_IGNORE_ERRORS` into the Hops command
- the test-helper's `HopsCLI`-object's `start`- & `build`-methods now take environment variables
- the test-helper's `HopsCLI`-object's `start`- & `build`-methods make their `stderr`-output available to the test environment

The changes to the `HopsCLI`-object are needed in order to be able to test the `doctor`. Let me know, if you'd rather see those changes in a separate PR. Personally, I'd cherry-pick them if we can't agree on the approach and thus need a few more iterations on the implementation.

Also note, that the `doctor`'s README isn't updated yet. I'd wait for this until we agreed on the API changes.

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/info/doctor/mixin.core.js
+++ b/packages/info/doctor/mixin.core.js
@@ -1,5 +1,6 @@
 const { async } = require('mixinable');
 const { Mixin } = require('hops-mixin');
+const deprecate = require('depd')('hops-doctor');
 const { internal: bootstrap } = require('hops-bootstrap');
 const { createDoctor } = require('.');
 
@@ -44,9 +45,13 @@ class DoctorMixin extends Mixin {
     return Promise.resolve().then(() => {
       doctor.getMode().then((mode) =>
         this.diagnose(doctor, mode).then((results) => {
-          doctor.collectResults(
-            ...[].concat(...results.filter((result) => result !== undefined))
-          );
+          const messages = results.filter(Boolean);
+          if (messages.length) {
+            deprecate(
+              'Returning messages from the diagnose-hook is deprecated. Please use "pushWarning" and "pushError" instead.'
+            );
+          }
+          doctor.collectResults(...[].concat(...messages));
           doctor.logResults(this.getLogger());
         })
       );
--- a/packages/info/doctor/mixin.core.js
+++ b/packages/info/doctor/mixin.core.js
@@ -6,11 +6,31 @@ const { createDoctor } = require('.');
 const { parallel } = async;
 const { validate, invariant } = bootstrap;
 
+const getMode = (argv) => {
+  const { _: cmd = [], production } = argv;
+  const command = cmd.join(' ');
+
+  switch (true) {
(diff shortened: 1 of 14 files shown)


packages/info/doctor/mixin.core.js
```diff
@@ -21,32 +42,45 @@ class DoctorMixin extends Mixin {
   bootstrap() {
     const { doctor } = this;
 
-    return this.diagnose(doctor).then((results) =>
-      doctor.collectResults(
-        ...[].concat(...results.filter((result) => result !== undefined))
-      )
-    );
+    return Promise.resolve().then(() => {
+      doctor.getMode().then((mode) =>
```
jhiode, line 46: Do we already have a use case for this feature?

packages/spec/helpers/test-helpers.js
```diff
@@ -29,6 +31,7 @@ const startServer = ({ cwd, command, argv = [] }) =>
     const args = [hopsBin, command].concat(argv);
     debug('Spawning server', [hopsBin, ...args].join(' '));
     const started = spawn(process.argv[0], args, {
+      env,
```
jhiode, line 34: Why do we need this? The child process has access to the same environment as the parent process. We would only need to set this if we want to change the environment.

herschel666: Setting `process.env.WHATEVER` before calling `HopsCLI.start(…)` didn't lead to the env var being present in the spawned process. Maybe it works, if the env var is passed into the initiating process (rather than fiddling with the `process.env`-object), but that wouldn't really work, because we can only specifiy env vars for the whole test suite if I'm not mistaken.

packages/info/doctor/mixin.core.js
herschel666: One example is the port, where no feedback is required if the app is in `build`-mode. And I'm pretty sure we find cases, where the decision on emitting an error or a warning depends on whether the app is in `develop`- or (production-)`serve`-mode.

jhiode (approved):

herschel666:
bors merge

herschel666:
bors retry

herschel666:
bors retry

herschel666:
blocked by #1377

ZauberNerd:
bors merge

jhiode (approved):

herschel666:
bors merge

PR #1360, opened 2020-09-24
chore: correct error name for deserialization

jhiode:
A small typo slipped through 😬

The other side is correct though:
https://github.com/xing/hops/blob/143089a668a541de2542a4c50924ed9df03fc716/packages/webpack/lib/utils/errors.js#L62


herschel666 (approved):

jhiode:
bors merge

PR #1366, opened 2020-09-29
Prepare master branch for next major

ZauberNerd:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


lerna.json
```diff
@@ -1,7 +1,7 @@
 {
   "npmClient": "yarn",
   "useWorkspaces": true,
-  "version": "13.0.0",
+  "version": "14.0.0-preview.0",
```
KnisterPeter, line 4: I would suggest `next` as tag and version name

KnisterPeter (approved):

lerna.json
jhiode: I think it makes sense to have a version here as we need that when we deploy the demo app. How about `14.0.0-nightly` or just  `14.0.0-master`?  `14.0.0-next` would also be fine.

KnisterPeter: That's what I meant. I would just suggest to take 'next' instead of 'preview' because I think that's more common.

package.json
```diff
@@ -33,6 +33,7 @@
     "lint": "eslint 'packages/**/*.js'",
     "prerelease": "test -n \"$GH_TOKEN\" || (echo 'GH_TOKEN missing'; exit 1)",
     "release": "lerna publish",
+    "release:preview": "yarn release --conventional-prerelease --preid preview",
```
herschel666, line 36: Do we actually plan to release a `next`-version? Or do we rather switch to `alpha` before doing the very first pre-release of Hops 14?

herschel666 (changes requested):
`packages/spec/integration/typescript-unit-testing/package.json:28` should either be pulled up to the latest version or removed altogether.

lerna.json
ZauberNerd: The tag name is already `next`: https://github.com/xing/hops/blob/master/lerna.json#L14 and I changed preview to nightly.

package.json
ZauberNerd: I would like to do some nightly releases from master - therefore I added this command.

herschel666 (approved):

PR #1404, opened 2020-11-03
feat(info): make logger available in server runtime

herschel666:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/info/logger/mixin.core.js
+++ b/packages/info/logger/mixin.core.js
@@ -14,6 +14,19 @@ class LogMixin extends Mixin {
     this.handleArguments(this.options);
   }
 
+  configureServer(app, middlewares, mode) {
+    if (mode === 'static') {
+      return;
+    }
+
+    middlewares.preinitial.unshift((_, res, next) => {
+      Object.assign(res.locals, {
+        logger: this.logger,
+      });
+      next();
+    });
+  }
+
   getLogger() {
     return this.logger;
   }
--- a/packages/info/logger/mixin.core.js
+++ b/packages/info/logger/mixin.core.js
@@ -21,7 +21,7 @@ class LogMixin extends Mixin {
 
     middlewares.preinitial.unshift((_, res, next) => {
       Object.assign(res.locals, {
-        logger: this.logger,
+        logger: res.locals.logger || this.logger,
       });
       next();
     });
--- a/packages/info/logger/mixin.server.js
+++ b/packages/info/logger/mixin.server.js
@@ -1,0 +1,25 @@
+const { sync } = require('mixinable');
+const { Mixin } = require('hops-mixin');
+const { internal: bootstrap } = require('hops-bootstrap');
+
+const { callable } = sync;
+const { validate, invariant } = bootstrap;
+
+class ServerLogMixin extends Mixin {
+  bootstrap(_, res) {
+    const { logger } = res.locals;
+    this.logger = logger;
+  }
+
+  getLogger() {
+    return this.logger;
+  }
+}
+
+ServerLogMixin.strategies = {
+  getLogger: validate(callable, ({ length }) => {
+    invariant(length === 0, 'getLogger(): Received unexpected argument(s)');
+  }),
+};
+
+module.exports = ServerLogMixin;


packages/info/logger/mixin.server.js
```diff
@@ -0,0 +1,35 @@
...
+
+class ServerLogMixin extends Mixin {
+  constructor(...args) {
+    super(...args);
+
+    const [{ name, rootDir }] = args;
+
+    this.logger = createLogger({
+      name: `${name}:runtime`,
+      _workspace: rootDir,
```
herschel666, line 17: The `_workspace`-value from the config is used to cut pathnames in error stacks. But this value isn't available here. So the workaround is to use the `rootDir`, which might lead to non-truncated pathnames in a setup comprised of workspaces, because those values differ in this setup.

I'm open for suggestions on how to do this better! 😃

packages/info/README.md
```diff
@@ -54,6 +54,29 @@ module.exports = class FooMixin extends Mixin {
...
+
+This utility hook defined by `hops-info` provides other mixins with a fully configured logger instance. This logger has three standard log methods (`info`, `warning` and `error`) that correspond to the lower three log levels described above. Additionally, it supports arbitrary log methods (e.g. `request`, `hint`, `foo`).
+
+```javascript
+const { Mixin } = require('hops-mixin');
+
+module.exports = class BarRuntimeMixin extends Mixin {
+  foo() {
+    if (typeof this.getLogger === 'function') {
+      const logger = this.getLogger();
```
jhiode, line 67:
```suggestion
    if (typeof this.getRuntimeLogger === 'function') {
      const logger = this.getRuntimeLogger();
```
Btw, why do we need to check if `this.getLogger` is a function?

jhiode (changes requested):

packages/info/README.md
herschel666: That's a very good question!

ZauberNerd: Because earlier not all apps had the `@untool/info` package installed.

packages/info/logger/mixin.server.js
```diff
@@ -0,0 +1,35 @@
+const { sync } = require('mixinable');
+const { Mixin } = require('hops-mixin');
+const { internal: bootstrap } = require('hops-bootstrap');
+const { createLogger } = require('.');
+
+const { callable } = sync;
+const { validate, invariant } = bootstrap;
+
+class ServerLogMixin extends Mixin {
```
ZauberNerd, line 9: Could you not just pass the already configured logger through?
Use the core mixin to register a middleware and add the logger to res.locals and then in the server mixin you use bootstrap to get the logger out of res.locals and have a `getLogger` hook that returns the logger or undefined (when called before `bootstrap`.

ZauberNerd (changes requested):

packages/info/logger/mixin.server.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  bootstrap(_, res) {
+    const { logger } = res.locals;
+    this.logger = logger;
+    this.logBuffer.flushBuffer(logger);
+    this.logBuffer = null;
+  }
+
+  getRuntimeLogger() {
```
ZauberNerd, line 47: I would prefer that the method has the same name as the core method `getLogger`

packages/info/logger/mixin.server.js
```diff
@@ -0,0 +1,61 @@
+const { sync } = require('mixinable');
+const { Mixin } = require('hops-mixin');
+const { internal: bootstrap } = require('hops-bootstrap');
+
+const { callable } = sync;
+const { validate, invariant } = bootstrap;
+
+class LogBuffer {
```
ZauberNerd, line 8: Where do we have the case, that we call `this.getLogger()` and the logger isn't there yet so we need a buffer?
I believe that shouldn't happen, because `bootstrap()` is the first hook we call https://github.com/xing/hops/blob/master/packages/react/render/mixin.server.js#L45 after `render()` https://github.com/xing/hops/blob/master/packages/react/render/index.js#L22

ZauberNerd: And in the `constructor()`s of other mixins we don't have access to the mixin instance yet, so we can't call hooks from there.

packages/info/logger/mixin.core.js
```diff
@@ -14,6 +14,19 @@ class LogMixin extends Mixin {
     this.handleArguments(this.options);
   }
 
+  configureServer(app, middlewares, mode) {
+    if (mode === 'static') {
+      return;
+    }
+
+    middlewares.preinitial.push((_, res, next) => {
```
ZauberNerd, line 22: Let's use `unshift` here, to put it first

packages/info/README.md
ZauberNerd: And I guess also sometimes you call a hook before the logger is initialized.

ZauberNerd (changes requested):

ZauberNerd (approved):

packages/info/logger/mixin.core.js
```diff
@@ -14,6 +14,19 @@ class LogMixin extends Mixin {
...
   }
 
+  configureServer(app, middlewares, mode) {
+    if (mode === 'static') {
+      return;
+    }
+
+    middlewares.preinitial.unshift((_, res, next) => {
+      Object.assign(res.locals, {
+        logger: this.logger,
```
jhiode, line 24: I think we need to create a new logger for each request.

jhiode (changes requested):

packages/info/logger/mixin.core.js
herschel666: In the context of this home-grown logger of ours? Or in the context of sending request-specific logs to Logjam? Because the latter requires additional plumbing anyways.

jhiode (approved):

herschel666:
bors merge

PR #1408, opened 2020-11-13
feat(webpack): add fast refresh plugin

x0s3:
## Current state

Hops currently uses HMR to update the frontend, which is great, but it may be better with the new fast refresh version 😄 

### What is introduced in this PR
In this PR we introduce a new flag `--fast-refresh` to enable the new _react-refresh_ and _react-refresh-webpack-plugin_ module that are currently used in React.

### Why we need to add it
Fast-refresh is the enhanced version of HMR (also previously known as Hot Reloading) that allows us to get near-instant feedback for changes in our React components, with `--fast-refresh` enabled, most edits should be visible within a second or two preserving React local state in function components (and Hooks) by default.

If we need to force the state to be reset and a component to be remounted we can add this anywhere in the file we are working on `// @refresh reset` this directive is local to the file, and instructs _react-refresh_ to remount components defined in that file on every edit.

Another interesting features that this plugin gives to us is **error resilience**:

- If we make a syntax error during a `--fast-refresh` session, we can fix it and save the file again. The new redbox will disappear. Modules with syntax errors are prevented from running, so we won't need to reload the app 😄.

- If we make a mistake that leads to a runtime error inside our component, the `--fast-refresh` session will also continue after we fix the error. In that case, React will remount our application using the updated code

- If we make a runtime error during the module initialization, the `--rast-refresh` session will continue once we fix the error. The redbox will disappear, and the module will be updated.

But there are some **small limitations**, two reasons why we might see local state being reset on every edit to a file:

- Local state is not preserved for class components (it includes hocs that returns classes instead of functional components)
- The module we're editing might have other exports in addition to a React component

### Screenshots and an example with hops/template-react

FastHopsRefresh

It's good to mention that since hops uses an older version of webpack `4.41.0` we cannot use the latest version for _react-refresh-webpack-plugin_ since they are using the new [(see full changelog here)](https://github.com/pmmmwh/react-refresh-webpack-plugin/releases/tag/v0.4.0) `module.hot.invalidate API` that webpack releases in `v4.43.0`, so until hops can use a new version of webpack we have to stay with `v0.3.3` otherwise it won't work.

## Checklist

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


packages/template-react/src/counter/index.js
```diff
@@ -1,29 +1,25 @@
-import React from 'react';
+import React, { useState } from 'react';
```
x0s3, line 1: update component to allow use `--fast-refresh` without problems, if for some reason this component was class based let me know and I can revert this change 👍

packages/webpack/mixins/fastRefresh/mixin.core.js
```diff
@@ -0,0 +1,28 @@
+const { Mixin } = require('hops-mixin');
+const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
+
+class ReactRefreshMixin extends Mixin {
+  configureBuild(webpackConfig, { jsLoaderConfig }, target) {
+    const { fastRefresh } = this.options;
+
+    if (target === 'develop' && fastRefresh) {
```
x0s3, line 8: Correct me if I'm wrong but with this conditional we only update frontend config right? only for frontend when we are developing and the flag is true.

Another thing that I want to ask if it make sense to have a new mixin (this one) or would be better to update only the `WebpackOptimizationsMixin`?

ZauberNerd:
> It's good to mention that since hops uses an older version of webpack `4.41.0` we cannot use the latest version for _react-refresh-webpack-plugin_ since they are using the new [(see full changelog here)](https://github.com/pmmmwh/react-refresh-webpack-plugin/releases/tag/v0.4.0) `module.hot.invalidate API` that webpack releases in `v4.43.0`, so until hops can use a new version of webpack we have to stay with `v0.3.3` otherwise it won't work.

You could upgrade webpack to `^4.43.0`, to ensure that the correct version gets installed.

packages/template-react/src/counter/index.js
ZauberNerd: There's no specific reason. It's just old code ;)

packages/webpack/mixins/fastRefresh/mixin.core.js
ZauberNerd: Yes, that's correct.
I think it would make sense to put this into the optimisations mixin, because there's a code path that removes all babel plugins when TypeScript is being used.

When hops-typescript is installed, we add the babel loader with all of its options too, to be able to do automatic polyfilling and to make sure that `importComponent` works: https://github.com/xing/hops/blob/master/packages/typescript/mixin.core.js#L31-L45
But in the optimisations mixin, we remove all babel plugins except for our own `importComponent`: https://github.com/xing/hops/blob/master/packages/webpack/mixins/optimizations/mixin.core.js#L74-L88, so I think it makes sense to add fast-refresh there too.

ZauberNerd (commented):
Overall this looks very good. I have just a few suggestions.

herschel666:
This PR got superseded by #1720. Thanks for your contribution, @x0s3!

PR #1448, opened 2020-12-04
WIP: implement circuit breaker

ZauberNerd:
This commit introduces a two-staged circuit breaker which will:
* disable react's `renderToString` if the process has high utilization
* quickly respond with a 503 if the process has even higher utilization

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/express/lib/circuit-breaker.js
+++ b/packages/express/lib/circuit-breaker.js
@@ -1,0 +1,47 @@
+const { performance } = require('perf_hooks');
+
+// TODO: figure out good values
+const upperThreshold = 0.9;
+const lowerThreshold = 0.7;
+const intervalMs = 1000;
+
+function createCircuitBreaker(app) {
+  let elu = performance.eventLoopUtilization();
+  let lastMeasurementMs = performance.now();
+
+  return (req, res, next) => {
+    // we don't want a new elu measurement for each incoming request, so we
+    // throttle it based on how much time passed since the last measurement
+    const nowMs = performance.now();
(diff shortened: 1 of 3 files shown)


packages/express/lib/circuit-breaker.js
```diff
@@ -0,0 +1,62 @@
+const { performance } = require('perf_hooks');
+
+// TODO: figure out good values
+const upperThreshold = 0.75;
+const lowerThreshold = 0.6;
```
ZauberNerd, line 5: These values aren't tested yet, they're just a gut feeling. We need to deploy a sample app to our production cluster to fine-tune these values.
We should probably also make them configurable and provide sensible defaults.

packages/express/lib/circuit-breaker.js
```diff
@@ -0,0 +1,62 @@
...
+
+function createCircuitBreaker(app) {
+  let elu = performance.eventLoopUtilization();
+  app.locals.breaker = 'closed';
+  app.locals.noSSR = false;
+
+  let last = performance.now();
+
+  return (req, res, next) => {
+    elu = performance.eventLoopUtilization();
```
ZauberNerd, line 15: Right now we're calculating the ELU per each request, I'm not sure, if we should try to implement some sort of sliding-window over some configurable amount of time or whether we should take it each request.
Also: I'm not entirely sure, if we maybe should pass in an older ELU value into the `eventLoopUtilization()` method, to calculate a delta between two values?

PR #1470, opened 2020-12-23
chore(spec): improve programmatic HopsCLI

herschel666:
First draft of the improved `HopsCLI`. As soon as we'll have agreed upon the API, this should either be refactored out of `packages/spec` into its own released package. Or `packages/spec` needs to be released (without all the tests obviously…).

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/spec/helpers/env.js
+++ b/packages/spec/helpers/env.js
@@ -3,8 +3,6 @@ const debug = require('debug')('hops-spec:env');
 const {
   isPuppeteerDisabled,
   launchPuppeteer,
-  startServer,
-  build,
   createWorkingDir,
 } = require('./test-helpers');
 
@@ -41,16 +39,6 @@ function isReactLifecycleWarning(warning) {
 
 function isIntolerableWarning(type, text) {
   return type === 'warning' && !isReactLifecycleWarning(text);
-}
-
-function getCommandModifications(args) {
-  const [maybeEnv, ...rest] = args;
-  const env =
-    typeof maybeEnv === 'object' && maybeEnv !== null ? maybeEnv : undefined;
-  return {
-    argv: env ? rest : args,
-    env,
-  };
 }
 
 class FixtureEnvironment extends NodeEnvironment {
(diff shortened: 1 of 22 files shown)


herschel666:
currently blocked by #1480

ZauberNerd (changes requested):
I liked how you commented and typed the new API :+1: 
But other than that, I'm not really happy with it. It's usage seems more complicated than before:
- cwd is always required and before we were able to use it directly from the Jest environment, while now we have to pass it in explicitly every time
- the builder pattern is overly complicated and does not add value in this case
Overall, this feels a bit over-engineered to me and I'd prefer to go back to something simpler :smile:

herschel666:
I agree with you concerning the `cwd`-value & I switched back to using the global one.

Regarding your concern, that it's overly complicated, I wonder when was the last time you've worked with `HopsCLI`? Have you seen in the red part of this PR's diff, how complicated it became, to put in env vars and CLI-flags? And how cumbersome it became to get `stderr` and `stdout` out of it? The new `HopsCLI` has straight-forward APIs for all of that.

So I guess we need more opinions here. 🤷‍♂️

KnisterPeter:
I do not have a strong opinion here, since I've never used the HopsCLI tool, but I would question the builder-style API here.
I think its not very easy to use.
My biggest problem here is mostly: I don't know/understand why we need to change this. Is it internal only? Who needs access to this? What are the limiting factors in the current implementation.

herschel666:
So `HopsCLI` used to be a fire-and-forget tool — run a command & don't care about it. But lately the requirements have been introduced, to set env variables and to retrieve the `stderr` of the process.

This is where things got messy — under the hood as well as in the public API.

Hence the overhaul. And since there's some kind of state involved, I decided to go with a `class`. Maybe I got carried away a little regarding the API design…

I will see whether I can come up with a less _classy_ approach.

herschel666:
Sh*t, forgot to delete the obsolete helper functions. 😖 Now the diff is complete.

herschel666:
superseded by #1481

PR #1473, opened 2020-12-29
fix(postcss): update Webpack's CSS compilation

herschel666:
I left a bunch of comments, that explain what all the changes are about.

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


packages/postcss/mixin.core.js
```diff
@@ -1,10 +1,6 @@
-// XXX: This file creates multiple side-effects
-require('./patch-postcss-modules-extract-imports');
```
herschel666, line 2: Doesn't seem to be necessary anymore, the resolution of the imports in `spec/integration/postcss` works fine without it.

packages/postcss/mixin.core.js
```diff
@@ -13,17 +9,46 @@ const getAssetPath = (...args) => trimSlashes(join(...args));
 
 const cssLoaderLocalOptions = {
   modules: {
+    auto: () => true,
```
herschel666, line 12: `css-loader` enables CSS Modules only for files, whose names match `/\.module\.css$/`, when `auto` is set to `true`. From what I observed, we need it enabled for all files, so we rather pass a function into this option, that always returns `true`.

packages/postcss/mixin.core.js
```diff
@@ -13,17 +9,46 @@ const getAssetPath = (...args) => trimSlashes(join(...args));
...
     localIdentName: '[folder]-[name]-[local]-[hash:8]',
+    exportLocalsConvention: 'camelCase',
   },
-  localsConvention: 'camelCase',
+  esModule: false,
```
herschel666, line 16: `esModule` is `true` by default for all the CSS-related Webpack loaders & plugins. Explicitly setting it to `false` throughout this file, to keep the status quo.

packages/postcss/mixin.core.js
```diff
@@ -13,17 +9,46 @@ const getAssetPath = (...args) => trimSlashes(join(...args));
...
   sourceMap: process.env.NODE_ENV !== 'production',
 };
 
 const cssLoaderGlobalOptions = {
   modules: false,
+  esModule: false,
   sourceMap: process.env.NODE_ENV !== 'production',
 };
 
+const getPostCssLoader = (browsers) => ({
```
herschel666, line 26: Extracting this loader config, because for the `node`-build to succeed, we need all the imports to be resolved properly. Therefore it is applied there, too.

Question's though whether we actually need `postcss-preset-env` for the `node`-build?!?

packages/postcss/mixin.core.js
```diff
@@ -93,14 +96,17 @@ class PostCSSMixin extends Mixin {
...
       allLoaderConfigs.indexOf(jsLoaderConfig),
       0,
-      getCSSLoaderConfig(this.config.browsers, ExtractCSSPlugin.loader)
-    );
-
-    webpackConfig.plugins.unshift(
-      new FilterWarningsPlugin({
-        exclude: /\[mini-css-extract-plugin\][^]*Conflicting order between:/,
+      getCSSLoaderConfig(this.config.browsers, {
+        loader: ExtractCSSPlugin.loader,
```
herschel666, line 100: `mini-css-extract-plugin`'s loader apparently needs a little configuration now to work as expected.

packages/postcss/mixin.core.js
```diff
@@ -111,6 +117,7 @@ class PostCSSMixin extends Mixin {
           this.config.assetPath,
           '[name]-[contenthash:12].css'
         ),
+        ignoreOrder: true,
```
herschel666, line 120: This replaces the application of the `FilterWarningsPlugin`. 🎉

packages/postcss/mixin.core.js
```diff
@@ -131,7 +138,15 @@ class PostCSSMixin extends Mixin {
     allLoaderConfigs.splice(
       allLoaderConfigs.indexOf(jsLoaderConfig),
       0,
-      getCSSLoaderConfig(this.config.browsers, require.resolve('style-loader'))
+      getCSSLoaderConfig(this.config.browsers, {
+        loader: require.resolve('style-loader'),
```
herschel666, line 142: Like the `mini-css-extract-plugin`'s loader, the `style-loader` now also needs a little configuration to work as expected.

packages/postcss/mixin.core.js
```diff
@@ -141,24 +156,36 @@ class PostCSSMixin extends Mixin {
...
+                ...cssLoaderGlobalOptions,
+                modules: {
+                  auto: false,
+                  exportOnlyLocals: true,
+                },
+                importLoaders: 1,
+              },
             },
-          },
+            getPostCssLoader(this.config.browsers),
```
herschel666, line 171: As mentioned above, `postcss-loader` is applied here too, to get the imports straight.

packages/postcss/package.json
```diff
@@ -18,20 +18,17 @@
...
+    "css-loader": "^5.0.1",
     "hops-mixin": "^14.0.0-nightly.4",
     "mini-css-extract-plugin": "^1.0.0",
-    "optimize-css-assets-webpack-plugin": "^5.0.3",
+    "optimize-css-assets-webpack-plugin": "^5.0.4",
     "pathifist": "^1.0.0",
-    "postcss-import": "^12.0.1",
-    "postcss-loader": "^3.0.0",
-    "postcss-modules-extract-imports": "^2.0.0",
+    "postcss": "^8.2.1",
```
herschel666, line 26: Added `postcss` here explicitly. This shouldn't be necessary since the `css-loader` brings it, too. But `postcss-preset-env` isn't `postcss@8`-ready, yet, and still comes along with `postcss@7`. So this dependency is in here to force the correct version. :feelsgood:

packages/postcss/mixin.core.js
ZauberNerd: I suppose not. In the Node.js build we need only to compute the exported names, so the import plugin should be enough.

packages/postcss/mixin.core.js
herschel666: Because `auto: true` enables CSS Modules for files, whose names match /\.module\.css$/. So we have to provide a custom matcher function, which basically enables it for every file — whatever their name is.

packages/postcss/mixin.core.js
herschel666: I like the idea of changing its value in the next major release.

packages/postcss/mixin.core.js
herschel666: Removed it for the Node-build.

packages/postcss/mixin.core.js
herschel666: Added another commit, that breakingly changes the value. We can omit that commit, when cherry-picking.

ZauberNerd (approved):

herschel666:
bors merge

herschel666:
bors merge

PR #1481, opened 2021-01-06
chore(spec): refactor HopsCLI

herschel666:
supersedes #1470 

Not that much changes here, but running the `start`-command got decoupled from retrieving the URL. And retrieving the URL got decoupled from retrieving the function for stopping the server.

Also the values, the `build`-command resolves to, now reflects what `exec` is yielding.

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/spec/helpers/test-helpers.js
+++ b/packages/spec/helpers/test-helpers.js
@@ -1,7 +1,6 @@
 const { spawn } = require('child_process');
 const path = require('path');
 const { promisify } = require('util');
-const exec = promisify(require('child_process').exec);
 const rimraf = promisify(require('rimraf'));
 
 const { copy } = require('fs-extra');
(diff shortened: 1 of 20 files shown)


packages/spec/helpers/test-helpers.js
```diff
@@ -13,60 +14,63 @@ const resolveFrom = require('resolve-from');
...
+    stderr += line;
+  });
 
-    started.on('error', (error) => reject(error));
+  started.on('close', (code) => {
+    debug('Server stopped. exitcode:', code);
+    stopServer = () => {};
   });
 
+  started.on('error', (error) => teardownPromise.reject(error));
```
KnisterPeter, line 69: There is one thing missing here. You should add this:

```js
process.on('exit', () => {
  stopServer();
});
```

It's not guaranteed by unix systems to stop spawned processes in all cases. Even though it is not a detached process. To be on the save side, we want to stop all spawned processes.

packages/spec/helpers/test-helpers.js
```diff
@@ -13,60 +14,67 @@ const resolveFrom = require('resolve-from');
...
+  const returnValue = { stdout: '', stderr: '' };
   debug('Starting', command);
   try {
-    return await exec(command, { env, cwd });
-  } catch (err) {
-    return { stderr: err.message };
+    const result = await exec(command, { env, cwd });
+    Object.assign(returnValue, result);
+  } catch (error) {
+    Object.assign(returnValue, { error });
```
ZauberNerd, line 23: Why do we return different objects here?
I think we should always return `{ stdout: string, stderr: string, error: null | Error }` right?
Because now we need to get `stderr` through `error.message`, which could be made simpler: https://github.com/xing/hops/pull/1481/files#diff-85085ff19f5129dd6e046735baf4dbd5949625be39a9956e7a33b936f0454908R19

packages/spec/helpers/test-helpers.js
```diff
@@ -13,60 +14,67 @@ const resolveFrom = require('resolve-from');
...
+
+  started.stderr.on('data', (data) => {
+    const line = data.toString('utf-8');
+    debug('stderr >', line);
+    stderr += line;
   });
 
+  started.on('close', (code) => {
+    debug('Server stopped. exitcode:', code);
+    stopServer = () => {};
```
ZauberNerd, line 66: We should reject `urlPromise` here, otherwise we'll run into a Jest timeout in the doctor/develop integration test.

ZauberNerd (commented):
Overall this looks fine. Just two small suggestions.

packages/spec/helpers/test-helpers.js
herschel666: `stderr` and `error` are different things. Even if we decide to always have an `error`-property on the resolved object, which defaults to `null`. That's what I meant with…

> align HopsCLI.build-API with child_process.exec-API

ZauberNerd (approved):

KnisterPeter (approved):
LGTM

herschel666:
bors merge

PR #1482, opened 2021-01-07
 feat(jest-environment): add package jest-environment-hops

herschel666:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


packages/spec/jest.setup.js
```diff
@@ -1 +1,20 @@
 jest.setTimeout(5 * 60 * 1000);
+
```
herschel666, line 2: I pulled the check of the console output out of `createPage`, because it needs to be configurable.

Another approach might be, to pass in a path to module, that exports an `isIntolerableWarning`-function, via the [`testEnvironmentOptions`](https://jestjs.io/docs/en/configuration#testenvironmentoptions-object), import that module inside the custom environment & thereby keep the actual checking encapsulated.

WDYT?

packages/jest-environment/README.md
```diff
@@ -0,0 +1,111 @@
...
+Builds the local Hops app into a temporary folder.
+
+```js
+const { stderr, error } = await HopsCLI.build('-p');
+
+expect(stderr).toBe('');
+expect(error).not.toBeDefined();
+```
+
+### `HopsCLI.start: (env?: NodeJS.ProcessEnv, ...cliArgs: string[]): Promise<{ getUrl: () => Promise<string>, stopServer: () => Promise<void> }>`
```
ZauberNerd, line 62: I think `HopsCLI.start` itself does not return a promise, right?

packages/jest-environment/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "puppeteer": "^5.5.0",
+    "resolve-from": "^5.0.0",
+    "rimraf": "^3.0.2",
+    "url-join": "^4.0.1"
+  },
+  "devDependencies": {
+    "jest-environment-node": "^26.6.2"
+  },
+  "peerDependencies": {
+    "jest-environment-node": "^26.6.2"
```
ZauberNerd, line 34: Why do we need the Jest environment to be a peer dependency?

packages/spec/jest.setup.js
ZauberNerd: I think it's ok for now until we have more usages and can find a better solution.

packages/spec/jest.setup.js
```diff
@@ -1 +1,20 @@
...
+    'Warning: componentWill[^\\s]+ has been renamed, and is not recommended for use. ' +
+      'See https://fb.me/react-async-component-lifecycle-hooks for details.'
+  );
+  return Boolean(warning.match(reReactLifecycleWarning));
+};
+
+const isIntolerableWarning = (type, text) =>
+  type === 'warning' && !isReactLifecycleWarning(text);
+
+global.handleConsoleOutput = (msg) => {
```
ZauberNerd, line 14: Not sure why, but for some reason adding a global here seems different to me than the globals that the Jest environment exposes and I'd prefer it not be a global, but I can't quite put my finger on why it it's different..

ZauberNerd (commented):
Overall I like it. I left a few comments.
Has this branch been rebased against current master to contain the refactored HopsCLI?

packages/jest-environment/package.json
herschel666: I thought I make consistent with all the other Jest-related packages being peer-dependencies.

packages/spec/jest.setup.js
herschel666: At the end of the day the globals, exposed by the Jest env, suck, too, in a way. 🤷‍♂️ 

We can export this function from a `helper.js` in the `spec`-folder. It doesn't need to be global.

packages/jest-environment/package.json
ZauberNerd: We only have those Jest dependencies as peer dependencies that the user actually needs in their own code.
For indirect dependencies that the user doesn't need to import, we have this in the Jest preset: https://github.com/xing/hops/blob/master/packages/jest-preset/jest-preset.js#L2-L18

herschel666: Yeah, I remember that one.

packages/spec/jest.setup.js
ZauberNerd: After thinking about this a bit more, I think the reason why this doesn't sit well with me is:
Providing globals through the Jest environment is part of their API contract (provided by `this.globals` instead of `global` - the Node global environment) and is considered acceptable (see `describe`, `it` and `expect` functions).
Furthermore: We have documented them in the package and on can easily find them by looking up the environment.
Once we start adding more globals from different places (such as a `jest.setup.js` file), we distribute the places where one has to look them up.
But as said in my comment above, this is mostly hairsplitting and probably some gut feeling.

herschel666: Fine for me. So what do you think about importing the function from a helper file instead?

ZauberNerd: Yes, sounds good :+1:

packages/jest-environment/README.md
```diff
@@ -0,0 +1,111 @@
...
+[![npm](https://img.shields.io/npm/v/jest-environment-hops.svg)](https://www.npmjs.com/package/jest-environment-hops)
+
+Please see the [main Hops Readme](../../DOCUMENTATION.md) for general information and a Getting Started Guide.
+
+> **This package is part of Hops's internal tooling. Hence, it doesn't provide an API contract. Breaking changes may be introduced even with patch releases. Use at your own risk.**
+
+## Installation
+
+```bash
+npm install --save-dev jest-environment-hops jest-environment-node
```
ZauberNerd, line 12: Remove `jest-environment-node` from the list

packages/jest-environment/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "type": "git",
+    "url": "https://github.com/xing/hops.git",
+    "directory": "packages/jest-environment"
+  },
+  "main": "index.js",
+  "dependencies": {
+    "debug": "^4.1.1",
+    "eprom": "^1.0.0",
+    "fs-extra": "^9.0.1",
+    "jest-environment-node": "^26.6.2",
```
ZauberNerd, line 23: Having this as a direct dependency might bring us trouble with updated Jest major versions again.
We should probably implement it the same way as we did here: https://github.com/xing/hops/blob/master/packages/jest-preset/jest-preset.js#L2-L18 or what do you think?

packages/jest-environment/index.js
```diff
@@ -1,22 +1,41 @@
+const colors = require('colors');
+const importFrom = require('import-from');
+const jestConfig = importFrom.silent(require.resolve('jest'), 'jest-config');
```
ZauberNerd, line 3: This should be `NodeEnvironment`, not `jest-config` as we don't need the config here, but the node environment ;)

packages/jest-environment/package.json
ZauberNerd: The dependency to `jest-environment-node` should be removed here.

packages/jest-environment/index.js
herschel666: Ah, sh*t, forgot that this ceremony wasn't only about checking the vesion. 😖 I'm a bit unfocused today…

ZauberNerd (approved):

herschel666:
bors merge

PR #1509, opened 2021-01-25
refactor: correct lodash import

KnisterPeter:
According to lodash docs (https://lodash.com/per-method-packages) the
per method packages shouldn't be used any longer.

The bootstrap code of hops is also bundled in the client and is
therefore directly related to the described problem.


ZauberNerd (approved):

ZauberNerd:
bors merge

KnisterPeter:
V12 is here: https://github.com/xing/hops/pull/1512

PR #1632, opened 2021-02-26
feat(webpack): introduce experimental esbuild transpilation

jhiode:
To test this just add `esbuild` and `esbuild-loader` as dependencies to `hops-template-redux`.

```
$ time -l yarn hops build -p
hops-template-react:info running 'build' in 'production' mode
hops-template-react:info bundling 'node' finished after 5.6s (2.93 MB)
- server.js (2.93 MB)
hops-template-react:info bundling 'build' finished after 12.9s (318 kB)
- hops-template-react-1-71d927b348e8.js (279 kB)
- hops-template-react-4531ec959528.js (38.2 kB)
- main-e94a689b8054.css (207 B)
✨  Done in 17.02s.
```

```
$ time -l yarn hops build -p --experimentalEsbuild
hops-template-react:info running 'build' in 'production' mode
hops-template-react:info bundling 'node' finished after 4.2s (2.59 MB)
- server.js (2.59 MB)
hops-template-react:info bundling 'build' finished after 4s (223 kB)
- hops-template-react-1-89661b912d7e.js (211 kB)
- hops-template-react-202879a799d9.js (11.7 kB)
- main-e94a689b8054.css (207 B)
✨  Done in 8.63s.
```

Memory consumption also goes down from ~340MB to ~240MB.


packages/webpack/mixins/esbuild/mixin.core.js
```diff
@@ -0,0 +1,85 @@
+const { Mixin } = require('hops-mixin');
+const importComponentTransform = require.resolve(
+  './importComponentTransform.js'
+);
+
+const isLoaderFor = (file) => (test) => {
```
herschel666, line 6: I'd use `fileType` here instead of `file` to make the purpose more obvious.

packages/webpack/lib/plugins/log.js
```diff
@@ -46,10 +46,9 @@ const formatWarning = (name, duration, assets, isRebuild) => {
 };
 
 const formatSuccess = (name, duration, assets, isRebuild) => {
-  const totalSize = Object.values(assets).reduce(
-    (sum, asset) => sum + asset.size,
-    0
-  );
+  const totalSize = Object.values(assets).reduce((sum, asset) => {
+    return sum + (asset.name.endsWith('.map') ? 0 : asset.size);
```
herschel666, line 50: This would be better readable with a `.filter(…)` before the `.reduce(…)`, instead of the ternary operator.

packages/webpack/mixins/esbuild/mixin.core.js
herschel666: I think it's fine to only handle `RegExp | Regexp[]` for now. But maybe it makes sense to warn, if it's a string or an object. So people with custom loader configurations know, that certain parts of their code (or the complete code) won't be bundled with esbuild.

jhiode: > I'd use fileType here instead of file to make the purpose more obvious.

The test runs against a file though (AFAICT)

>  So people with custom loader configurations know, that certain parts of their code (or the complete code) won't be bundled with esbuild.

I think it is obvious that esbuild will only transpile JavaScript and TypeScipt files. Getting warnings for e.g. `.css` because I am using the `--experimentalEsbuild` flag will create confusion IMO.

herschel666: > The test runs against a file though (AFAICT)

Yes, but the `file`-argument represents a pattern — a string ending with either `.js` or `.ts`. There won't ever be passed a file path into that function during the runtime of Webpack. That's why the name is not optimal.

jhiode: I've applied your second suggestion but I would like to leave the `file` thingy as is.

packages/webpack/mixins/esbuild/mixin.core.js
```diff
@@ -0,0 +1,89 @@
...
+
+const isLoaderFor = (extention) => (test) => {
+  if (Array.isArray(test)) {
+    return test.some(isLoaderFor(extention));
+  } else if (typeof test === 'object' && typeof test.test === 'function') {
+    return test.test(extention);
+  } else if (typeof test === 'function') {
+    return test(extention) === true;
+  } else if (typeof test === 'string') {
+    return test === extention;
```
herschel666, line 14: If the value of `test` is of type `String`, the condition is fulfilled when the filepath either starts with the `test`-string or equals the `test`-string. So I think you should write a warning in this case, that the test is not sufficient, and either do `test.endsWith(extension)` or simply return `false`.

Btw: what do you think of…?

```js
switch (true) {
  case Array.isArray(test):
    return test.some(isLoaderFor(extention));
  case typeof test === 'object' && typeof test.test === 'function':
    return test.test(extention);
  case typeof test === 'function':
    return test(extention) === true;
  case typeof test === 'string':
    return test.endsWith(extention);
  default:
    return false;
}
```

jhiode:
Jest is also mostly working. We have one Flow test that isn't working, which is fine I guess and there seems to be another import syntax that is hard to match via regex. Not sure what we do about that.

The syntax I am talking about:
https://github.com/xing/hops/blob/5ba1ffc3a9a37b42a15ac4848f167c6d9cdc06c6/packages/jest-preset/helpers/__tests__/fixture/code-02.txt#L3

packages/webpack/mixins/esbuild/importComponentTransform.js
```diff
@@ -0,0 +1,10 @@
+const regex = /importComponent\s*\(\s*\(\)\s+=>\s+import\(\s*'([^']+)'\s*\)\s*\)/g;
+
+function transformImportComponent(source) {
+  return source.replace(
+    regex,
+    "importComponent({ load: () => import('$1'), moduleId: require.resolveWeak('$1') })"
+  );
+}
+
+module.exports = transformImportComponent;
```
herschel666, line 10: I think this should be named something with `loader` instead of `transform`. Otherwise I find it pretty confusing.

herschel666 (commented):
Regarding the import component loader: couldn't we leverage acorn when we find ocurrences of `importComponent(` in the code and then basically replicating thte respective babel loader code?

Or we check whether there's a second parameter passed into `importComponent`, and if there is, issue an error telling the user, that Hops's esbuild-feature currently only supports default exports in `importComponent`. 🤷‍♂️

packages/webpack/mixins/esbuild/importComponentTransform.js
jhiode: On the other hand, it isn't loading anything but it is actually transforming a string 🤷‍♂️

jhiode:
Btw, in ESBuild there are four different loaders: `js`, `jsx`, `ts`, `tsx`. For webpack I'm just using `jsx` for `jsx?` and `tsx` for `tsx?`. For JavaScript that shouldn't be an issue as any JS file is also a valid JSX file. However, that is [not true for TS and TSX](https://esbuild.github.io/content-types/#ts-vs-tsx).
So for the Jest loader I just kept the default to use the extension to detect which loader should be used. For that, I had to rename a lot of (internal) files though (see https://github.com/xing/hops/pull/1632/commits/d75c5d77aaac249331a82f9a91c7a2d082d7b57d).

I think I should align that to the WebPack loader as well, but that would require everyone who wants to use this esbuild feature to possibly fix their file extensions.

packages/webpack/mixins/esbuild/importComponentTransform.js
herschel666: I didn't tell Webpack to call their transorms `loader`. 😅

herschel666:
> I think I should align that to the WebPack loader as well, but that would require everyone who wants to use this esbuild feature to possibly fix their file extensions.

I think it would be fine to introduce this requirement for an experimental feature coming with a major update. But if we want to go further down the esbuild road, we'll eventually have to introduce that requirement to everybody. I personally think that's appropriate. Do we all agree on that?

ZauberNerd:
> Regarding the import component loader: couldn't we leverage acorn when we find ocurrences of `importComponent(` in the code and then basically replicating thte respective babel loader code?

I took a stab at that a few days ago, unfortunately, the webpack parser APIs are not so easy and also pretty much not documented.
I suppose it should look something like that: https://github.com/xing/hops/commit/f81cb4af87c6e9765bb602f9aec2b385d1418c2a, but I'm not sure :laughing: 
There's only 11 results on GitHub for `parser.hooks.call.for`: https://grep.app/search?q=parser.hooks.call.for :(

I guess it'd be easier to continue with the RegExp based solution if we want to finish it in time for Hops 14.

> Or we check whether there's a second parameter passed into importComponent, and if there is, issue an error telling the user, that Hops's esbuild-feature currently only supports default exports in importComponent. :man_shrugging:

I'd prefer, if we can handle named imports, I'll take a look at the regex, so see if I can figure out a way.

ZauberNerd:
Another thing to keep in mind: There seems to be some incompatibility when using esbuild artifacts in webpack: https://github.com/evanw/esbuild/issues/758 (see also PR 2855 in crate). We should evaluate if that is also the case here.

packages/react/build/mixin.core.js
```diff
@@ -2,24 +2,36 @@ const { Mixin } = require('hops-mixin');
...
     jsLoaderConfig.test.push(/\.jsx$/);
 
-    jsLoaderConfig.options.presets.push(require.resolve('@babel/preset-react'));
-
-    jsLoaderConfig.options.plugins.push(
-      require.resolve('@babel/plugin-transform-flow-strip-types'),
-      require.resolve('@babel/plugin-proposal-class-properties')
-    );
+    if (experimentalEsbuild) {
+      jsLoaderConfig.use[0].options.loader = 'jsx';
```
ZauberNerd, line 12: I'm not quite sure, what `use[0]` should be here. I tried logging it with: `DEBUG=hops:webpack:config:develop DEBUG_DEPTH=99` but it seems `use` is not an array?
Maybe it would make sense to make this more robust, like we do in other situations, to find a loader by its name.

packages/styled-components/mixin.core.js
```diff
@@ -2,6 +2,15 @@ const { Mixin } = require('hops-mixin');
 
 class StyledComponentsMixin extends Mixin {
   configureBuild(webpackConfig, { jsLoaderConfig }, target) {
+    const { experimentalEsbuild } = this.options;
+
+    if (experimentalEsbuild) {
+      console.warn(
+        'The experimental esbuild transpilation does not support styled components yet!'
```
ZauberNerd, line 9: It doesn't support it for SSR, right? But it would still give useful output in dev mode?
Maybe change this warning a bit, to indicate that it can still be used in dev mode.

ZauberNerd (commented):
Documentation is missing. For testing you need to set `USE_EXPERIMENTAL_ESBUILD` and otherwise you use `--experimental-esbuild` or the env var. It would be good to have that documented ;)

packages/webpack/mixins/esbuild/mixin.core.js
```diff
@@ -0,0 +1,56 @@
...
+    const { experimentalEsbuild } = this.options;
+
+    if (!experimentalEsbuild) {
+      return;
+    }
+
+    // NOTE: We don't want to ship esbuild, so we just assume it is installed
+    // when someone wants to use it.
+    // eslint-disable-next-line node/no-missing-require, node/no-extraneous-require
+    const { ESBuildPlugin, ESBuildMinifyPlugin } = require('esbuild-loader');
```
herschel666, line 14: The [`ESBuildPlugin` can be removed now.](https://github.com/privatenumber/esbuild-loader/releases/tag/v2.10.0)

packages/typescript/mixin.core.js
```diff
@@ -28,29 +28,78 @@ const getCompilerOptions = (ts, rootPath) => {
...
-      test: /\.tsx?$/,
-      exclude,
-      use: [
+    webpackConfig.resolve.extensions.push('.ts', '.tsx');
+
+    if (experimentalEsbuild) {
+      const {
+        include,
+        exclude,
+        use: [{ loader, options }, ...loaders],
```
herschel666, line 39:
```suggestion
        use: [{ loader: esbuildLoader, options }, ...loaders],
```

To make it a little more clear what's going on here.

herschel666 (changes requested):

packages/react/build/mixin.core.js
jhiode: In the `.experimentalEsbuild` case this is always an array because we have the import component transformation/loader here as well.

packages/styled-components/mixin.core.js
jhiode: I haven't really looked into it so I'm not sure yet what works and what doesn't.

packages/webpack/mixins/esbuild/mixin.core.js
jhiode: Hm, but we would still need to boilerplate code for the webpack API adapter, right? Why is it even deprecated?

packages/spec/integration/esbuild-typescript/yarn.lock
```diff
@@ -0,0 +1,5741 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ZauberNerd, line 1: How did this file end up here?

ZauberNerd, line 1: How did this file end up here?

packages/spec/integration/esbuild-typescript/package.json
```diff
@@ -0,0 +1,32 @@
+{
```
ZauberNerd, line 1: Should we also run a jest test in this integration test, to demonstrate how it works?

packages/spec/integration/esbuild-typescript/yarn.lock
jhiode: Oops...

systemboogie (commented):
Even if that command-line flag has an "experimental" in its name I wonder: Shouldn't Hops print a warning/info saying that it is experimental and some things may not work as expected (apart from styled components) and that a production build is created differently?

ZauberNerd (approved):

herschel666 (approved):

jhiode:
bors merge

PR #1655, opened 2021-03-12
Run eslint and prettier as jest-runner, update npm scripts

ZauberNerd:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


package.json
```diff
@@ -157,9 +157,40 @@
...
+        ],
+        "testMatch": [
+          "<rootDir>/packages/**/*.js"
+        ]
+      },
+      {
+        "displayName": "prettier",
+        "runner": "prettier",
+        "preset": "jest-runner-prettier",
+        "testPathIgnorePatterns": [
```
herschel666, line 180: Since this runner apparently [does not respect `.prettierignore`](https://github.com/keplersj/jest-runner-prettier/issues/87), we should add `package.json` and `package-lock.json` here.

herschel666 (changes requested):

package.json
ZauberNerd: We don't use npm, so we don't have a `package-lock.json`, do you think we should still add it?
And I think having the `package.json` formatted by prettier is ok nowadays. We excluded it initially because a `yarn add` or `npm install` would result in a differently formatted file, but as far as I can see this is no longer an issue, right?

package.json
```diff
@@ -157,9 +157,40 @@
...
+    "watchPlugins": [
+      "jest-runner-eslint/watch-fix",
+      "jest-watch-select-projects",
+      "jest-watch-typeahead/filename",
+      "jest-watch-typeahead/testname"
+    ],
     "projects": [
       "<rootDir>/packages/!(spec)",
-      "<rootDir>/packages/spec/integration/*"
+      "<rootDir>/packages/spec/integration/*",
```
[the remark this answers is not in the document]
ZauberNerd: - renaming `test` to `unit` is fine for me
- renaming `template` to `unit` is fine for me
- I'm not sure though, how we can provide the config on the top-level, because through the glob expression each one of them becomes a "project" entry. So in order to provide them here, we'd need to add a "project" entry for each package, I guess? Or have you found a different way?

package.json
herschel666: Hm, haven't checked that in a while. I'm happy to not have `package.json` ignored by Prettier. But I think we should still ignore `package-lock.json`, because npm is the package manager referenced in the docs and contributors might use it.

package.json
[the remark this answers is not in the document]
herschel666: Oh, didn't know of that implication of the glob expression here. It's a little unfortunate to have that project name in the templates's `package.json` files. But I don't have a better idea…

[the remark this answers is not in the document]
ZauberNerd: I've renamed `test` and `templates` to `unit`

packages/spec/integration/jest-preset/__tests__/setup/global-css.js
```diff
@@ -1,2 +1,2 @@
-// eslint-disable-next-line import/no-unresolved
+// eslint-disable-next-line import/no-unresolved, node/no-missing-import
```
herschel666, line 1: The filename is misspelled, it's missing a `t`.

Also, when we install [eslint-import-resolver-webpack](npm.im/eslint-import-resolver-webpack), the linter knows how to deal with resource queries.

packages/template-graphql/.npmignore
```diff
@@ -0,0 +1,2 @@
+__tests__/
```
herschel666, line 1: We should create more `.npmignore`s! 👍

package.json
```diff
@@ -144,22 +138,59 @@
...
   ],
   "workspaces": [
     "packages/*"
   ],
   "jest": {
+    "collectCoverage": true,
+    "coverageProvider": "v8",
+    "testPathIgnorePatterns": [
+      "/node_modules/",
+      "<rootDir>/packages/spec/.tmp/",
```
herschel666, line 156: We could as well ignore all the `.tmp`-folders, like the eslint config is doing it.

herschel666 (changes requested):
I get the warning…

```
You provided values for --selectProjects but a project does not have a name.
Set displayName in the config of all projects in order to disable this warning.
```

…when running the linter only. Shouldn't every project have a display name?!? 🤔

package.json
```diff
@@ -144,22 +138,59 @@
...
+      "jest-watch-typeahead/testname"
+    ],
     "projects": [
       "<rootDir>/packages/!(spec)",
-      "<rootDir>/packages/spec/integration/*"
+      "<rootDir>/packages/spec/integration/*",
+      {
+        "displayName": "lint",
+        "runner": "eslint",
+        "testPathIgnorePatterns": [
```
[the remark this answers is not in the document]
ZauberNerd: Unfortunately not :(

packages/spec/integration/jest-preset/__tests__/setup/global-css.js
ZauberNerd: That plugin relies on a webpack config.

herschel666: It can work with defaults, when setting…

```yaml
settings:
  import/resolver: webpack
```

This fixed this specific issue & didn't produce any new ones. I don't know, though, whether this will cause any trouble in the future…

herschel666 (approved):

KnisterPeter (approved):

ZauberNerd:
bors merge

PR #1665, opened 2021-03-17
fix(postcss): use module type "commonjs"

herschel666:
This reverts the breaking change introduced with [`5f4ff80`](5f4ff80#diff-f6674fc8d197d1faca718594bf4931825ac8c5b1ba7aa51d523d3e0fd0a50633).

It also renders the previously introduced preset option `namedExport`
obsolete.

Switching to ESM has been a bold move and it turned out, that it not
only has the potential to break existing Hops apps through their own
CSS. But also through third-party packages, that themselves import CSS
Modules in their code. In unfortunate cases they may be using
namespace imports, while they're actually not able to be switched to
named exports, because of reserved keywords in their CSS classnames or
duplicate class names after camelCasification.

In those situations, the third-party package would not be usable.

However, by sticking with CommonJS as a module type for CSS Modules, the
import syntax is not that important. Both ways — default imports and
namespace imports — work fine in conjunction with the Webpack-based
bundling.

<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/postcss/README.md
+++ b/packages/postcss/README.md
@@ -79,37 +79,7 @@ For more information on that topic, please have a look at [the documentation of 
 
 #### Preset Options
 
-| Name | Type | Default | Required | Description |
-| --- | --- | --- | --- | --- |
-| `postcss.namedExport` | `Boolean` | `false` | _no_ | Whether CSS Modules do a named exports of class names rather than a default export of the class name object. |
-
-##### `postcss.namedExport`
-
-This option enables you to opt into CSS Modules exporting named exports, instead of doing a default export of the object, that holds all the class names.
-
-To enable it, set the following option in your Hops config.
-
-```json
-"hops": {
-  "postcss": {
-    "namedExport": true
-  }
-}
-```
-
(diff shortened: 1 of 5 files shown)


ZauberNerd (approved):
Generally this is fine. I wonder though, what would happen to the changelog if you were to use a regular `revert` commit (such as is being generated by `git revert ...`).
Because as I understand it, we now will have the breaking change notice in the generated changelog, right?
Maybe you could do a small test with `yarn lerna version --no-push --no-git-tag-version` to check how the changelog would be generated in both cases?

herschel666:
@ZauberNerd I changed the commit message. This produces a changelog entry like…

```markdown
### Reverts

* use module type "commonjs" instead of "esm" ([cd07037](https://github.com/xing/hops/commit/cd07037c669328ce2d541cb0718a5ef802f0be75))
* ```

…and also removes the entry concerning the commit containing the breaking change.

ZauberNerd (approved):

herschel666:
bors merge

PR #1727, opened 2021-04-14
test(spec): check experimental esbuild feature

herschel666:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>


yarn.lock
```diff
@@ -6035,11 +6035,6 @@ esbuild@^0.10.2:
   resolved "https://registry.yarnpkg.com/esbuild/-/esbuild-0.10.2.tgz#caa65a8f3096d547d89159918039df6c5c6c90be"
   integrity sha512-/5vsZD7wTJJHC3yNXLUjXNvUDwqwNoIMvFvLd9tcDQ9el5l13pspYm3yufavjIeYvNtAbo+6N/6uoWx9dGA6ug==
 
-esbuild@^0.9.6:
```
ZauberNerd, line 6038: Don't we need esbuild as a dependency?

packages/spec/integration/esbuild/__tests__/production.js
```diff
@@ -0,0 +1,35 @@
+const { handleConsoleOutput } = require('../../../helpers');
+
+describe('esbuild - production', () => {
+  let url, page, getInnerText;
+
+  beforeAll(async () => {
+    const { getUrl } = HopsCLI.start('--production', '--experimental-esbuild');
```
ZauberNerd, line 7: Oh, is a production build a valid use-case? Or do we want to disallow this?

yarn.lock
herschel666: `esbuild-loader` brings its own `esbuild`.

packages/spec/integration/esbuild/__tests__/production.js
herschel666: Good question! Might make sense. But we need documentation of that feature then.

ZauberNerd: Yes, documentation makes sense. Do you want to add it in this PR or create a separate one?
I know that @jhiode also offered to help with esbuild related tasks.

herschel666: I'll do it. Internally as well.

packages/webpack/README.md
```diff
@@ -82,7 +82,17 @@ This feature is enabled by default and can be disabled via the `--no-parallel-bu
...
 ##### `--fast-dev` _experimental_
 
-Using the experimental `--fast-dev` option will disable automatic polyfilling and transpiling of all `node_modules` files through babel to enable faster development times. This will lead to a different bundle being created than in production mode and will not work on all browsers (modern browsers only). Use with caution and report any bugs you may encounter.
+Using the experimental `--fast-dev` option will disable automatic polyfilling and transpiling of all `node_modules` files through babel to enable faster development times. This will lead to a different bundle being created than in production mode and will not work in all browsers (modern browsers only). Use with caution and report any bugs you may encounter.
+
+**DO NOT USE THIS MODE FOR QA OR PRODUCTION**
+
+##### `--experimental-esbuild` _experimental_
+
+Using the experimental `--experimental-esbuild` option will replace the `babel-loader` of Hops's internal Webpack config with the `esbuild-loader`. While ESBuild is significantly faster then Babel, it's still early-stage and might lead to unexpected results.
```
jhiode, line 91:
```suggestion
Using the experimental `--experimental-esbuild` option will replace the `babel-loader` of Hops's internal Webpack config with the `esbuild-loader`. While ESBuild is significantly faster than Babel, it's still early-stage and might lead to unexpected results.
```

packages/spec/package.json
```diff
@@ -37,6 +37,8 @@
     "@types/styled-components": "^5.1.8",
     "create-hops-app": "15.0.0-nightly.1",
     "cross-fetch": "^3.0.4",
+    "esbuild-jest": "^0.5.0",
+    "esbuild-loader": "^2.10.0",
```
jhiode, line 41: I would still add the `esbuild` dependency here as well as `esbuild-loader` might get fewer releases than `esbuild` and otherwise we would only get `esbuild` updates via the `yarn.lock` file maintenance - which feels wrong to me.

packages/spec/integration/esbuild/__tests__/production.js
jhiode: I think we should still allow esbuild mode to be used with production, I don't think that we should judge that. The experimental `--fast-build` mode also works with `--production`.

packages/spec/package.json
herschel666: I checked it & `esbuild-loader` uses the `esbuild` it brings with it. So there's no use in adding your own `esbuild`. Also the `esbuild`, that's a transient dependency of `esbuild-loader` is guaranteed to work. Which might not be the case with the `esbuild` you add yourself to the project.

packages/spec/integration/esbuild/__tests__/production.js
herschel666: Agreed. For the OS Hops this should be fine.

packages/spec/package.json
jhiode: Ok, you are right. We are only using/importing `esbuild-loader`. So it makes sense then. 👍

DOCUMENTATION.md
```diff
@@ -239,6 +239,7 @@ _The production flag (`--production` or `-p`) is a shortcut for setting the envi
 - `npm start` starts the development server with hot module reloading, etc.
 - `npm start -- --fast-refresh` (**experimental**) starts the development server with react-refresh, etc.
 - `npm start -- --fast-dev` (**experimental**) starts the development server with some optimizations for faster builds (_warning:_ this will lead to different bundles than in production, use with caution).
+- `npm start -- --experimental-esbuild` (**experimental**) starts the development server with [ESBuild]()https://esbuild.github.io/ instead of [Babel](https://babeljs.io/) (_warning:_ this will lead to different bundles than in production, use with caution). [Read here how to set it up.](packages/webpack#--experimental-esbuild-experimental)
```
jhiode, line 242: > this will lead to different bundles than in production, use with caution

That's not quite true as the esbuild mode could also be used in production.

herschel666: Good point!

packages/spec/integration/esbuild/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "jest": {
+    "displayName": "integration",
+    "testEnvironment": "jest-environment-hops",
+    "setupFilesAfterEnv": [
+      "../../jest.setup.js"
+    ]
+  },
+  "scripts": {
+    "start": "hops start",
+    "build": "hops build"
```
jhiode, line 20:
```suggestion
    "start": "hops start --experimental-esbuild",
    "build": "hops build --experimental-esbuild"
```
It's a small thing though and not really required.

jhiode (dismissed):
🎉

jhiode (approved):

herschel666:
bors merge

herschel666:
bors retry

ZauberNerd:
bors retry

PR #1730, opened 2021-04-15
fix: correctly initialize import component context

### LINKED ISSUE #1729: ImportComponent issue: modules.push is not a function
## This is my intent (choose one)

- [x] I want to report a bug
- [ ] I want to request a feature or change
- [ ] I want to provide or change a feature

## The problem

If I use `importComponent` I'll get the error `modules.push is not a function`.
This is caused by the object put into the `Provider` by default.
When using the hops bootstrap, this is overwritten by an array, which is correct.

## Proposed solution

Put `undefined` in the context in the first place:

https://github.com/xing/hops/blob/1922cbfdd380a44b7aef5e71d16028c77824a5a9/packages/react/import-component/context.js#L3

Additionally (and maybe better), this code should behave identically with empty array and undefined.

https://github.com/xing/hops/blob/1922cbfdd380a44b7aef5e71d16028c77824a5a9/packages/react/import-component/index.js#L69-L78

KnisterPeter:
The import component context should be initialized with undefined,
as this is the expected behavior of the Import Component.

Closes #1729

### DIFF
--- a/packages/react/import-component/context.js
+++ b/packages/react/import-component/context.js
@@ -1,6 +1,6 @@
 import { createContext } from 'react';
 
-const context = createContext({});
+const context = createContext(undefined);
 
 context.displayName = 'Hops.ImportComponent';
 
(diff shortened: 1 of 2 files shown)


jhiode (approved):

KnisterPeter:
bors merge

PR #1878, opened 2021-06-21
chore: allow usage of hops with node 16+

KnisterPeter:
This change add node 16.x to the allowed engines.
It specifically allows to install hops in an environment with node 16 and yarn without flags.
There are no compatibility changes in this PR.


herschel666 (approved):
❤️

ZauberNerd (approved):
Looks good. But should we also add Node.js 16 to the test matrix? https://github.com/xing/hops/blob/master/.github/workflows/node-ci.yml#L19

herschel666 (approved):

ZauberNerd (approved):

PR #1889, opened 2021-06-24
feat: add hops-msw package

ZauberNerd:
<details>

| Syntax | Description |
| --- | --- |
| bors try- | Cancel a try |

</details>

### DIFF
--- a/packages/jest-preset/jest-preset.js
+++ b/packages/jest-preset/jest-preset.js
@@ -17,6 +17,8 @@ if (Number(jestMajorVersion) < 26) {
   );
 }
 
+const enableMockServiceWorker = process.env.ENABLE_MSW === 'true';
+
 const useEsbuild = process.env.USE_EXPERIMENTAL_ESBUILD === 'true';
 const jsTransform = useEsbuild
   ? require.resolve('./transforms/esbuild.js')
@@ -52,9 +54,12 @@ module.exports = {
     '^.+\\.(gql|graphql)$': require.resolve('./transforms/graphql.js'),
   },
   transformIgnorePatterns: [],
-  setupFiles: useEsbuild
-    ? []
-    : [require.resolve('regenerator-runtime/runtime')],
+  setupFiles: [
+    useEsbuild ? undefined : require.resolve('regenerator-runtime/runtime'),
+  ].filter(Boolean),
+  setupFilesAfterEnv: [
+    enableMockServiceWorker ? require.resolve('./setup-msw.js') : undefined,
+  ].filter(Boolean),
   // fixes: https://github.com/facebook/jest/issues/6766
   testURL: 'http://localhost',
   testEnvironment: 'jsdom',
--- a/packages/msw/README.md
+++ b/packages/msw/README.md
@@ -1,0 +1,49 @@
+# `hops-msw`
+
+[![npm](https://img.shields.io/npm/v/hops-msw.svg)](https://www.npmjs.com/package/hops-msw)
+
+**Please see the [main Hops Readme](../../DOCUMENTATION.md) for general information and a Getting Started Guide.**
+
+This is a [preset for Hops](../../DOCUMENTATION.md#presets) that can be used to set up a [Mock Service Worker](https://www.npmjs.com/package/msw) for your Hops app. It allows to mock REST and GraphQL APIs in unit and integration tests.
+
+This package (together with `jest-preset-hops`) takes care of setting up the MSW mock server already, so if you use it for unit-testing, you do not need to do this yourself.
+
+**Example:**
+
+```javascript
+import { render, waitFor } from '@testing-library/react';
+import { mockServer, graphql } from 'hops-msw';
+import { withApolloTestProvider } from 'hops-react-apollo';
+
+it('loads graphql data', async () => {
+  mockServer.use(
+    graphql.query('search', (req, res, ctx) => {
+      return res(
+        ctx.data({
+          __typename: 'Query',
+          user: {
+            __typename: 'User',
+            name: 'Something',
+          },
+        })
+      );
+    })
+  );
+
+  const { getByText } = render(withApolloTestProvider(<HomeWithData />));
+
+  await waitFor(() => getByText('Something'));
+});
+```
+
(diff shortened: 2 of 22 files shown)


packages/msw/mixin.core.js
```diff
@@ -0,0 +1,110 @@
...
+  return content;
+};
+
+module.exports = class MswMixin extends Mixin {
+  configureBuild(webpackConfig) {
+    if (process.env.ENABLE_MOCKS !== 'true') {
+      webpackConfig.resolve = webpackConfig.resolve || {};
+      webpackConfig.resolve.alias = webpackConfig.resolve.alias || {};
+      webpackConfig.resolve.alias['hops-msw'] = false;
+      return;
```
ZauberNerd, line 29: Why do we need this? With this change we create different bundles based on the env var `ENABLE_MOCKS`, which means we cannot use the production artifact to run the tests.
Since we're using a dynamic `import()` in the bootstrap method of the browser mixin, we shouldn't have bigger bundles anyway.

packages/msw/mixin.browser.js
```diff
@@ -0,0 +1,49 @@
...
+        ctx.status(200),
+        typeof data === 'undefined' ? undefined : ctx.json(data)
+      )
+    );
+  }
+};
+
+class MswMixin extends Mixin {
+  async bootstrap() {
+    if (process.env.ENABLE_MOCKS === 'true') {
```
ZauberNerd, line 25: Here we should probably use `this.config` instead and map the env var `ENABLE_MOCKS` to a config value, so that we can test any bundle.

packages/msw/mixin.core.js
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    const { EnvironmentPlugin } = require('webpack');
+
+    webpackConfig.plugins.push(
+      new EnvironmentPlugin({
+        ENABLE_MOCKS: process.env.ENABLE_MOCKS ? 'true' : undefined,
+      })
+    );
```
ZauberNerd, line 38: We won't need this, if we use a config value to map the env var.

packages/jest-preset/jest-preset.js
```diff
@@ -17,6 +17,8 @@ if (Number(jestMajorVersion) < 26) {
   );
 }
 
+const enableMocks = process.env.ENABLE_MOCKS === 'true';
```
jhiode, line 20: I would like to have a better name here as `mocks` is a general term and it's not obvious that we are talking about service worker code that mocks specific network reqeusts.

packages/msw/mixin.browser.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  if (type === 'rest') {
+    return rest[method](identifier, (_, res, ctx) =>
+      res(
+        ctx.status(200),
+        typeof data === 'undefined' ? undefined : ctx.json(data)
+      )
+    );
+  }
```
jhiode, line 20: The `graphql[method]` and `rest[method]` accesses are looking a little bit dangerous. I think it could be easy to write a wrong mock that would cause a runtime error here. I think having a helpful error message instead would be great in such a case.

The two `if`s also look weird / could be put in a `if/else-if/else`-clause to also detect invalid mock types.

```js
  if (type === 'graphql') {
    // ...
  } else if (type === 'rest') {
    // ...
  } else {
    // warn about unhandled mock request
  }
```

packages/msw/mixin.core.js
```diff
@@ -0,0 +1,93 @@
...
+                mockServer.use(
+                  rest[method](identifier, (_, res, ctx) =>
+                    res(
+                      ctx.status(200),
+                      typeof data === 'undefined' ? undefined : ctx.json(data)
+                    )
+                  )
+                );
+              }
+            });
```
jhiode, line 75: Same here

packages/msw/mixin.core.js
```diff
@@ -0,0 +1,93 @@
...
+  ]);
+
+  const content = readFileSync(join(dir, 'mockServiceWorker.js'));
+
+  return content;
+};
+
+module.exports = class MswMixin extends Mixin {
+  configureServer(_, middlewares) {
+    if (process.env.ENABLE_MOCKS === 'true') {
```
jhiode, line 25: It's minor but in my opinion very helpful to reduce the indentations and keep the code simpler to read.
```suggestion
    if (process.env.ENABLE_MOCKS !== 'true') return;
```

jhiode (commented):
It looks great so far 🎉 
I've left some comments and suggestions though.

packages/jest-preset/jest-preset.js
ZauberNerd: We took that name from crate, where it is also `ENABLE_MOCKS`.

jhiode: I know, but is it required that it has the same name? I'm happy to keep the name if you think it makes sense.
But to me, it isn't very descriptive of what it is doing though. And I would also try to prevent leaking internal company "requirements" into Hops - whenever possible.

Wow about `ENABLE_MSW` or `ENABLE_SW_MOCKS`? Not sure how expressive `SW` and `MSW` are though.

packages/msw/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Installation
+
+Add this preset to your existing Hops React project:
+
+```bash
+npm install --save hops-msw
+```
+
+In order to use it for unit-testing, you also need to install the [`jest-preset-hops`](../jest-preset-hops/README.md) package.
```
jhiode, line 49: I think we should add a usage description and mention the environment variable here

packages/jest-preset/jest-preset.js
herschel666: We'll rename it to `ENABLE_MSW` to make it more dedicated to the `msw` thingie. But the naming should suffice then. If users write that env var, they likely have informed about its existence in beforehand.

herschel666: @jhiode done; ptal again 🙏

jhiode (approved):

herschel666:
bors merge

herschel666:
bors retry

PR #2010, opened 2021-11-04
feat: use swc for import component transpilation (in case of esbuild)

jhiode:
This replaces the RegEx-based `importComponent` transformer for esbuild. 

There is still one difference to the babel version, SWC removes all comments from the result and there is no way around it yet.


ZauberNerd:
We've replaced the loader with babel in: https://github.com/xing/hops/pull/2070

